PR #136, opened 2018-02-05
Add @Px annotation where applicable

Jawnnypoo:
There might be some places that were missed, but I think since ktx depends on the support annotations package, it would make sense to specify these parameters are expected to be pixel values. 

Any thoughts on this?


JakeWharton:
Yes we use these "type" annotations quite liberally. These were just an oversight. Thanks!

PR #137, opened 2018-02-05
Feature/simple extensions

abbath0767:
add The simplest and most frequently used extensions:

- visibility state
- set `VISIBLE`, `INVISIBLE` and `GONE` states
- doAfterDraw - extension for perform action when the global layout state or the visibility of views within the view tree changes


JakeWharton:
Please break these up into separate PRs so they can be discussed individually.

PR #140, opened 2018-02-05
Add another common extensions

dewarder:
The purpose of this PR to discuss and add extensions which my colleagues and I use on a daily basis during development. 

I've added extensions for `TextView` and `SeekBar` which allows to attach/set listeners in hande way:

```kotlin
val editText: EditText = ...
editText.doAfterChanged { s: Editable -> }
editText.doBeforeChanged { s, start, count, after ->  }
editText.doOnChanged { s, start, before, count ->  }
    
val seekBar: SeekBar = ...
seekBar.doOnProgressChanged { progress, fromUser ->  }
seekBar.doOnStartTracking {  }
seekBar.doOnStopTracking {  }
```

Another things which I would like to add are version of `post`/`postDelayed` which accept `View` as receiver in block. It is helpful when you want to apply some changes from another thread:
```kotlin
val refreshLayout: SwipeRefreshLayout = ...
refreshLayout.postApply {
    isRefreshing = false
    background = ...
}
```

One more extension is about inflating views directly from `Context` or `ViewGroup`. It is super handy when you working with `RecyclerView`:

```kotlin
override fun onCreateViewHolder(parent: ViewGroup, viewType: Int) =
        MyViewHolder(parent.inflate(R.layout.my_layout))
```
And the last simple extension is controlling `View` visibility by boolean parameter rather than `View.GONE/View.VISIBLE`:

```kotlin
val view: View = ...
view.isVisible = true
```

Concerns:
- SeekBar listener extensions override each other and can't be used simultaneously
- `postApply/postDelayedApply` returns `Runnable` which inconsistent with Kotlin's `apply` where it returns receiver

In case if you also find this extensions useful I can proceed working with documentations or tests if it is needed from my side.

### DIFF
--- a/src/main/java/androidx/content/Context.kt
+++ b/src/main/java/androidx/content/Context.kt
@@ -19,9 +19,13 @@ package androidx.content
 import android.content.Context
 import android.content.res.TypedArray
 import android.support.annotation.AttrRes
+import android.support.annotation.LayoutRes
 import android.support.annotation.RequiresApi
 import android.support.annotation.StyleRes
 import android.util.AttributeSet
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
 
 /**
  * Return the handle to a system-level service by class.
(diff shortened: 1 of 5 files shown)


JakeWharton:
Please break these out into separate PRs for discussion. Also you'll need tests for each.

PR #145, opened 2018-02-05
Fixes Duration.kt methods requiring API level 260

fgsguedes:
This PR removes the invalid API level requirement of 260 and uses the appropriate one required by the android framework one used in the extension.

Instrumented tests were run locally on a Nexus 6P with 8.1.0.


JakeWharton:
260? That's like Android version ZZZZZZZZZZZZZ.

fgsguedes:
Just realized that merging this will leave the `api/current.txt` not reflecting the real current state of the library api.

Is that a change I should've done?

JakeWharton:
Yes. `current.txt` is what's on `master.

PR #146, opened 2018-02-05
Add simple function for view visibilities

abbath0767:
Add simple methods for check and set visibility state of view

### DIFF
--- a/src/main/java/androidx/view/View.kt
+++ b/src/main/java/androidx/view/View.kt
@@ -203,4 +203,4 @@ fun View.isGone() = visibility == View.GONE
  *
  *  @return true if visibility for current view is INVISIBLE
  */
-fun View.isHide() = visibility == View.INVISIBLE
+fun View.isHidden() = visibility == View.INVISIBLE
--- a/src/main/java/androidx/view/View.kt
+++ b/src/main/java/androidx/view/View.kt
@@ -196,7 +196,7 @@ inline fun View.gone() {
 /** Checks visibility state of view. If it is VISIBLE then return true. */
 fun View.isVisible() = visibility == View.VISIBLE
 
-/** Checks visibility state of view. If it is INVISIBLE then return true. */ */
+/** Checks visibility state of view. If it is INVISIBLE then return true. */
 fun View.isHidden() = visibility == View.INVISIBLE
 
 /** Checks visibility state of view. If it is GONE then return true. */
--- a/src/main/java/androidx/view/View.kt
+++ b/src/main/java/androidx/view/View.kt
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
     }
     return Bitmap.createBitmap(width, height, config).applyCanvas(::draw)
 }
+
+/** Show current view */
+fun View.show() {
+    visibility = View.VISIBLE
+}
+
+/** Hide current view */
+fun View.hide() {
+    visibility = View.INVISIBLE
+}
+
+/** Gone current view */
+fun View.gone() {
+    visibility = View.GONE
+}
+
+/** Return state of current view
+ *
+ *  @return true if visibility for current view is VISIBLE
+ */
+fun View.isVisible() = visibility == View.VISIBLE
+
+/** Return state of current view
+ *
+ *  @return true if visibility for current view is GONE
+ */
+fun View.isGone() = visibility == View.GONE
+
+/** Return state of current view
+ *
+ *  @return true if visibility for current view is INVISIBLE
+ */
+fun View.isHide() = visibility == View.INVISIBLE
--- a/src/main/java/androidx/view/View.kt
+++ b/src/main/java/androidx/view/View.kt
@@ -172,35 +172,32 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
     return Bitmap.createBitmap(width, height, config).applyCanvas(::draw)
 }
 
-/** Show current view */
-fun View.show() {
+/** Makes view visible.
+ *  It's equivalent to calling view.visibility = View.VISIBLE.
+ * */
+inline fun View.show() {
     visibility = View.VISIBLE
 }
 
-/** Hide current view */
-fun View.hide() {
+/** Makes view invisible.
+ *  It's equivalent to calling view.visibility = View.INVISIBLE.
+ * */
+inline fun View.hide() {
     visibility = View.INVISIBLE
 }
 
-/** Gone current view */
-fun View.gone() {
+/** Makes view gone.
+ *  It's equivalent to calling view.visibility = View.GONE.
+ * */
+inline fun View.gone() {
     visibility = View.GONE
 }
 
-/** Return state of current view
- *
- *  @return true if visibility for current view is VISIBLE
- */
+/** Checks visibility state of view. If it is VISIBLE then return true. */
 fun View.isVisible() = visibility == View.VISIBLE
 
-/** Return state of current view
- *
- *  @return true if visibility for current view is GONE
- */
-fun View.isGone() = visibility == View.GONE
-
-/** Return state of current view
- *
- *  @return true if visibility for current view is INVISIBLE
- */
+/** Checks visibility state of view. If it is INVISIBLE then return true. */ */
 fun View.isHidden() = visibility == View.INVISIBLE
+
+/** Checks visibility state of view. If it is GONE then return true. */
+fun View.isGone() = visibility == View.GONE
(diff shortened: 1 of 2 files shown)


src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+ *
+ *  @return true if visibility for current view is GONE
+ */
+fun View.isGone() = visibility == View.GONE
+
+/** Return state of current view
+ *
+ *  @return true if visibility for current view is INVISIBLE
+ */
+fun View.isHide() = visibility == View.INVISIBLE
```
MGaetan89: Rather `isHidden`?

abbath0767: your right! fix!

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
     }
     return Bitmap.createBitmap(width, height, config).applyCanvas(::draw)
 }
+
+/** Show current view */
```
JakeWharton: Missing period. [Comments should be sentences.](https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html).

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+fun View.show() {
```
JakeWharton: `inline`

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+    visibility = View.VISIBLE
+}
+
+/** Hide current view */
```
JakeWharton: period

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+fun View.hide() {
```
JakeWharton: `inline`

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+    visibility = View.INVISIBLE
+}
+
+/** Gone current view */
```
JakeWharton: period

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+fun View.gone() {
```
JakeWharton: `inline`

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+    visibility = View.GONE
+}
+
+/** Return state of current view
```
JakeWharton: A view has lots of state...

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+ *
+ *  @return true if visibility for current view is VISIBLE
```
JakeWharton: Don't say the same thing twice. Just
```kotlin
/** Returns true if the view is visible. */
```
gets the point across.

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+ */
+fun View.isVisible() = visibility == View.VISIBLE
+
+/** Return state of current view
+ *
+ *  @return true if visibility for current view is GONE
```
JakeWharton: Same as above

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+ */
+fun View.isGone() = visibility == View.GONE
+
+/** Return state of current view
+ *
+ *  @return true if visibility for current view is INVISIBLE
```
JakeWharton: Same as above

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+/** Gone current view */
+fun View.gone() {
+    visibility = View.GONE
+}
+
+/** Return state of current view
+ *
+ *  @return true if visibility for current view is VISIBLE
+ */
+fun View.isVisible() = visibility == View.VISIBLE
```
JakeWharton: `inline`

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+
+/** Return state of current view
+ *
+ *  @return true if visibility for current view is GONE
+ */
+fun View.isGone() = visibility == View.GONE
```
JakeWharton: `inline`

src/main/java/androidx/view/View.kt
```diff
@@ -171,3 +171,36 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+
+/** Return state of current view
+ *
+ *  @return true if visibility for current view is INVISIBLE
+ */
+fun View.isHidden() = visibility == View.INVISIBLE
```
JakeWharton: `inline`

src/main/java/androidx/view/View.kt
romainguy: I'm a bit ambivalent of the naming of this method. show and hide are both verbs but gone isn't. I wonder if we could find a better name for consistency.

AOrobator: How about `remove()`? It doesn't actually get removed from the view hierarchy, but it gets treated like it.  But the most straightforward names would be 

`View.setVisible()`
`View.setInvisible()`
`View.setGone()`

ouattararomuald: .Net has Visibility Collapsed ([Visibility Enumeration](https://msdn.microsoft.com/en-us/library/ms590101.aspx)). Maybe it won't be intuitive but what about `collapse` ?

pchmielowski: IMO, `Collapsed` is way better name than `Gone`. However, unless we can time travel and do Refactor->Rename on the `GONE`constant ten year ago, we shouldn't introduce second name for the same concept.

src/main/java/androidx/view/View.kt
jshvarts: @JakeWharton Is there a good rule when to inline?

JakeWharton: In this case you're aliasing to a public API and it doesn't require any allocation which is always a good candidate

src/main/java/androidx/view/View.kt
jshvarts: What about 
    
    /** Makes view visible. */

src/main/java/androidx/view/View.kt
Edward608: Maybe `View.goAway()`/`View.getOut()`?

src/main/java/androidx/view/View.kt
abbath0767: @JakeWharton thx for link. I will read the article and try to leave comments better for next commit. Prior to that, simply, did not use comments except in rare cases

src/main/java/androidx/view/View.kt
abbath0767: I understand that perhaps the name is not the best, but it is semantically close to the already declared meaning - GONE, to which we are all accustomed. I 100% agree with the opinion of @pchmielowski

arnis71: Maybe name them simply like so?
```
View.visible()
View.invisible()
View.gone()
```

abbath0767:
@JakeWharton thx. I understand you. Soon I will correct and I will be defined with a method name

src/main/java/androidx/view/View.kt
burnoo: How about
```kotlin
View.makeVisible()
View.makeInvisible()
View.makeGone()
```

michaldrabik: Here is another one: 
```
View.show()
View.hide()
View.begone() //Still a verb unlike View.gone()
```

src/main/java/androidx/view/View.kt
romainguy: To be extra clear you could add a sentence saying it's equivalent to calling `view.visibility = View.VISIBLE`.

src/main/java/androidx/view/View.kt
romainguy: Since none of the suggestions really work, let's keep `gone()` and add this to the long list of APIs we regret :)

PR #151, opened 2018-02-05
Add TextView listeners extensions

hluhovskyi:
Add extensions for TextView listeners which can be used in the following way:

```kotlin
val editText: EditText = ...
editText.doAfterChanged { s: Editable -> }
editText.doBeforeChanged { s, start, count, after ->  }
editText.doOnChanged { s, start, before, count ->  }
```


src/main/java/androidx/widget/TextView.kt
```diff
@@ -0,0 +1,81 @@
...
+
+        override fun beforeTextChanged(s: CharSequence, start: Int, count: Int, after: Int) {
+            beforeChanged?.invoke(s, start, count, after)
+        }
+
+        override fun onTextChanged(s: CharSequence, start: Int, before: Int, count: Int) {
+            onChanged?.invoke(s, start, before, count)
+        }
+    }
+    addTextChangedListener(watcher)
```
pchmielowski: What do you think about following?

`return object : TextWatcher { /* ... */ }.also { addTextChangedListener(it) }`

dewarder: For me variant with local variable looks clearer and it also follows convention which is used in `Animator`

Edward608: Which one do you think is better? `((Editable) -> Unit)? = null` or `(Editable) -> Unit = {}`?

JakeWharton: That is unrelated to these comments, but null.

src/main/java/androidx/widget/TextView.kt
```diff
@@ -0,0 +1,81 @@
...
+    beforeChanged: (s: CharSequence, start: Int, count: Int, after: Int) -> Unit
+): TextWatcher = addTextChangedListener(beforeChanged = beforeChanged)
+
+/**
+ * Add an action which will be invoked when text of this TextView has been changed.
+ *
+ * @return the [TextWatcher] added to the TextView
+ * @see [TextWatcher.onTextChanged]
+ */
+fun TextView.doOnChanged(
```
romainguy: I feel these methods should be called `doOnTextChanged`, `doBeforeTextChanged` and `doAfterTextChanged` for clarity and consistency with `addTextChangedListener` and the methods on `TextWatcher`.

dewarder: Makes sense. I am going to update PR soon.

hluhovskyi:
@JakeWharton @romainguy updated PR to use `androidx.core` package

EzraLopez:
What work is left here? Anyway I could help?

JakeWharton:
Text watcher-based listeners have been merged in http://r.android.com/720364

Issue #147, opened 2018-02-05
TextView.setCompoundDrawable... with default arguments


tomaszrykala:
eg.

```kotlin
@RequiresApi(Build.VERSION_CODES.JELLY_BEAN_MR1)
fun TextView.setCompoundDrawablesRelative(
        left: Drawable? = null,
        top: Drawable? = null,
        right: Drawable? = null,
        bottom: Drawable? = null
) {
    setCompoundDrawablesRelative(left, top, right, bottom)
}
// same for the other, similar methods
```

JakeWharton:
Name won't work. Extensions cannot override members.

tomaszrykala:
True, thanks. So replace `set` with `update`, eg. `fun TextView.updateCompoundDrawablesRelative(...)` ?

JakeWharton:
Yeah. And that's unfortunate. We should do non-relative too.

PRs welcome.

tomaszrykala:
Will do, thanks.

fredy-mederos:
The compound drawables extensions could be "var" extensions like this?

```kotlin
var TextView.drawableTop: Drawable?
    get() = compoundDrawables[1]
    set(value) {
        setCompoundDrawables(
            compoundDrawables[0],
            value,
            compoundDrawables[2],
            compoundDrawables[3]
        )
    }
```
Thanks

consp1racy:
@fredy-mederos That's `var TextView.compoundDrawableTop`.

consp1racy:
@tomaszrykala `null` is a valid argument for setting compound drawables. Perhaps we should use a marker object.

```
private object DoNotChange : Drawable() {
    override fun draw(canvas: Canvas) {
    }

    override fun setAlpha(alpha: Int) {
    }

    override fun getOpacity(): Int = PixelFormat.TRANSPARENT

    override fun setColorFilter(colorFilter: ColorFilter?) {
    }
}

fun TextView.updateCompoundDrawablesWithIntrinsicBounds(
        left: Drawable? = DoNotChange,
        top: Drawable? = DoNotChange,
        right: Drawable? = DoNotChange,
        bottom: Drawable? = DoNotChange
) {
    val compoundDrawables = compoundDrawables
    val newLeft = if (left != DoNotChange) left else compoundDrawables[0]
    val newTop = if (top != DoNotChange) top else compoundDrawables[1]
    val newRight = if (right != DoNotChange) right else compoundDrawables[2]
[...]

PR #159, opened 2018-02-05
Add TextView.setCompoundDrawable...

tomaszrykala:
Addresses https://github.com/android/android-ktx/issues/147


src/androidTest/java/androidx/view/TextViewTest.kt
```diff
@@ -0,0 +1,103 @@
...
+    private val context = InstrumentationRegistry.getContext()
+    private val resources = context.resources
+    private val textView = TextView(context)
+
+    private val resourceId = R.drawable.btn_default
+    private val drawable: BitmapDrawable
+        get() {
+            return BitmapDrawable(resources,
+                    createBitmap(10, 10).apply {
+                        eraseColor(Color.RED)
```
romainguy: Erasing it to a color isn't useful in this case since you never test what the drawable draws.

src/androidTest/java/androidx/view/TextViewTest.kt
```diff
@@ -0,0 +1,103 @@
...
+import android.graphics.Color
+import android.graphics.drawable.BitmapDrawable
+import android.support.test.InstrumentationRegistry
+import android.widget.TextView
+import androidx.graphics.createBitmap
+import org.junit.Assert
+import org.junit.Test
+
+class TextViewTest {
+
```
romainguy: Remove empty line

tomaszrykala:
Comments addressed, thanks @romainguy

src/androidTest/java/androidx/view/TextViewTest.kt
```diff
@@ -0,0 +1,99 @@
...
+    private val drawable: BitmapDrawable
+        get() {
+            return BitmapDrawable(resources, createBitmap(10, 10))
+        }
+
+    @Test
+    fun updateCompoundDrawables() {
+        textView.updateCompoundDrawables(start = drawable, end = drawable)
+
+        Assert.assertNotNull(textView.compoundDrawables[0])
```
JakeWharton: import these functions directly, please

src/androidTest/java/androidx/view/TextViewTest.kt
```diff
@@ -0,0 +1,99 @@
...
+    private val textView = TextView(context)
+    private val resourceId = R.drawable.btn_default
+    private val drawable: BitmapDrawable
+        get() {
+            return BitmapDrawable(resources, createBitmap(10, 10))
+        }
+
+    @Test
+    fun updateCompoundDrawables() {
+        textView.updateCompoundDrawables(start = drawable, end = drawable)
```
JakeWharton: I would prefer if you created drawables directly in these test methods and then used `assertSame` to assert that the correct drawable is set to the correct position. You can use `ColorDrawable` for an easy one to instantiate.

src/androidTest/java/androidx/view/TextViewTest.kt
```diff
@@ -0,0 +1,99 @@
...
+import android.widget.TextView
+import androidx.graphics.createBitmap
+import org.junit.Assert
+import org.junit.Test
+
+class TextViewTest {
+    private val context = InstrumentationRegistry.getContext()
+    private val resources = context.resources
+    private val textView = TextView(context)
+    private val resourceId = R.drawable.btn_default
```
JakeWharton: Inline this to its use sites. It doesn't add value as a property here.

src/androidTest/java/androidx/view/TextViewTest.kt
```diff
@@ -0,0 +1,99 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.view
+
+import android.R
```
JakeWharton: Normally `android.R` is never imported and always used as a fully-qualified reference.

src/main/java/androidx/view/TextView.kt
```diff
@@ -0,0 +1,116 @@
...
+ * Updates this TextView's Drawables. This version of the method allows using named parameters
+ * to just set one or more Drawables.
+ *
+ * @see TextView.setCompoundDrawables
+ */
+fun TextView.updateCompoundDrawables(
+        start: Drawable? = null,
+        top: Drawable? = null,
+        end: Drawable? = null,
+        bottom: Drawable? = null
```
JakeWharton: nit: 4 space indentation for these lines. I probably need to update the style files...

src/androidTest/java/androidx/view/TextViewTest.kt
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (C) 2017 The Android Open Source Project
```
romainguy: Oops sorry I missed this, the copyright year should be 2018

src/main/java/androidx/view/TextView.kt
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright (C) 2017 The Android Open Source Project
```
romainguy: Same here

src/androidTest/java/androidx/view/TextViewTest.kt
tomaszrykala: Sure, I copied it from other classes . I can correct this in my PR and all other in another or in the same?

src/main/java/androidx/view/TextView.kt
```diff
@@ -0,0 +1,116 @@
...
+    setCompoundDrawablesWithIntrinsicBounds(start, top, end, bottom)
+}
+
+/**
+ * Updates this TextView's Drawables. This version of the method allows using named parameters
+ * to just set one or more Drawables.
+ *
+ * @see TextView.setCompoundDrawablesRelative
+ */
+@RequiresApi(Build.VERSION_CODES.JELLY_BEAN_MR1)
```
JakeWharton: nit: we use the integer values directly in these annotations because they render better in the `current.txt`.

src/main/java/androidx/view/TextView.kt
tomaszrykala: Again, correct just my classes and create another PR for the rest?

src/androidTest/java/androidx/view/TextViewTest.kt
romainguy: Existing files that say 2017 are correct, you should only worry about new files.

src/androidTest/java/androidx/view/TextViewTest.kt
romainguy: Actually you can even use 0xffff0000.toDrawable() :)

src/main/java/androidx/view/TextView.kt
romainguy: We should update the style files, I made the same mistake a few times.

src/androidTest/java/androidx/view/TextViewTest.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.Assert
+import org.junit.Test
+
+class TextViewTest {
+    private val context = InstrumentationRegistry.getContext()
+    private val resources = context.resources
+    private val textView = TextView(context)
+    private val resourceId = R.drawable.btn_default
+    private val drawable: BitmapDrawable
+        get() {
```
jshvarts: Why not 

    get() = BitmapDrawable(resources, createBitmap(10, 10))

szymanskip:
@tomaszrykala shouldn't those extensions be placed under `androidx.widget` package?

romainguy:
Yes, good catch. This should be in `androidx.widget`.

tomaszrykala:
Thanks @szymanskip . Corrected @romainguy .

tomaszrykala:
@romainguy Is there anything still missing in this PR?

romainguy:
Sorry, I completely forgot about this PR, my bad! Looking at it again, I think this extension should work like ViewGroup.updateMargins. For instance if calling:

```
udpateCompoundDrawables(left=whatever, right=whatever)
```

Then `top` and `bottom` should remain unchanged instead of becoming `null`. What do you think?

tomaszrykala:
@romainguy I could try that, although my original intention for this change was to remove the boilerplate of additional drawables to set when you really want to set one, via default arguments.

What is we have two sets of `updateCompoundDrawable...` methods : 
- updateCompoundDrawable... - behaving like the ones in this PR
- updateCompoundDrawablePreserving... - that preserves any existing values

Option two: allow additional `Boolean` argument, that would toggle this preserve/overwrite behaviour, eg.

```kotlin
fun TextView.updateCompoundDrawables(
    start: Drawable? = null,
    top: Drawable? = null,
    end: Drawable? = null,
    bottom: Drawable? = null,
    keepExisting: Boolean = false
) {
    // if (keepExisting) ...
    setCompoundDrawables(start, top, end, bottom)
}
```

JakeWharton:
Neither of those behaviors are intuitive for a method named "update". I agree that this method should only alter the drawables which are supplied. This is also solves your case of only setting a subset, as they default to not present anyway.

tomaszrykala:
It's easy enough to do this with methods that take drawables as arguments eg:
```kotlin
fun TextView.updateCompoundDrawables(
    start: Drawable? = compoundDrawables[0],
    top: Drawable? = compoundDrawables[1],
    end: Drawable? = compoundDrawables[2],
    bottom: Drawable? = compoundDrawables[3]
) {
    setCompoundDrawables(start, top, end, bottom)
}
```

but how to do this with methods that take `@DrawableRes Int`s in? , meaning how do I ensure that drawables aren't being overwritten? Call the methods that take in Drawables internally instead? In code like the follwing the `getDrawable()` calls are cause `@Deprecated` warnings, and theme one that takes Theme as an argument is from API 21.
```kotlin
/**
 * Updates this TextView's Drawables. This version of the method allows using named parameters
 * to just set one or more Drawables. Use 0 if you do not want a Drawable there.
 *
 * @see TextView.setCompoundDrawablesWithIntrinsicBounds
 */
fun TextView.updateCompoundDrawablesWithIntrinsicBounds(
        @DrawableRes start: Int = -1,
        @DrawableRes top: Int = -1,
        @DrawableRes end: Int = -1,
        @DrawableRes bottom: Int = -1
) {
    setCompoundDrawablesWithIntrinsicBounds(
            if (start == -1) compoundDrawables[0] else if (start == 0) null else resources.getDrawable(start),
            if (top == -1) compoundDrawables[1] else if (top == 0) null else resources.getDrawable(top),
            if (end == -1) compoundDrawables[2] else if (end == 0) null else resources.getDrawable(end),
            if (bottom == -1) compoundDrawables[3] else if (bottom == 0) null else resources.getDrawable(bottom)
    )

```

JakeWharton:
Yep. That's what `TextView` does except we'll use `ContextCompat.getDrawable`.

tomaszrykala:
Thanks @JakeWharton . I kept thinking of `DrawableCompat`.

tomaszrykala:
Done @romainguy @JakeWharton

tomaszrykala:
Any more work required?

src/main/java/androidx/widget/TextView.kt
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Updates this TextView's Drawables. This version of the method allows using named parameters
+ * to just set one or more Drawables.
+ *
+ * @see TextView.setCompoundDrawablesRelative
+ */
+@RequiresApi(17)
+fun TextView.updateCompoundDrawablesRelative(
+    start: Drawable? = compoundDrawables[0],
```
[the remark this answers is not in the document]
tomaszrykala: That's true about `compoundDrawables` vs `...Relative`, thanks. I've corrected this, and I suppose it would have been hard for my tests to pick that difference up as all they do is compare the instances of the Drawables.

With regard to the repeated creations of the arrays I don't know how could we combat this, as both relative and non-relative calls create the arrays by accessing the package-private `mDrawables` instance, which I can't access directly.

src/main/java/androidx/widget/TextView.kt
```diff
@@ -0,0 +1,133 @@
...
+import android.widget.TextView
+
+/**
+ * Updates this TextView's Drawables. This version of the method allows using named parameters
+ * to just set one or more Drawables.
+ *
+ * @see TextView.setCompoundDrawables
+ */
+fun TextView.updateCompoundDrawables(
+    start: Drawable? = compoundDrawables[0],
```
cyrilmottier: I wanted to collaborate to android-ktx on `TextView` and after digging a bit in the GitHub project, I found you PR. Nice work 👍 

The discussions you have here are really interesting and I actually ended up having the exact same questions. The feedback @romainguy gave is true. Using `compoundDrawables` 4 times to get the default value will possibly involve creating several arrays (one per call to `compoundDrawables`).

I ended up doing so:

```kotlin
private val drawableSentinel = DrawableContainer()

fun TextView.updateCompoundDrawablesWithIntrinsicBounds(
        left: Drawable? = drawableSentinel,
        top: Drawable? = drawableSentinel,
        right: Drawable? = drawableSentinel,
        bottom: Drawable? = drawableSentinel
) {
    val drawables = lazy(LazyThreadSafetyMode.NONE) { compoundDrawables }
    setCompoundDrawablesWithIntrinsicBounds(
            if (left === drawableSentinel) drawables.value[0] else left,
            if (top === drawableSentinel) drawables.value[1] else top,
            if (right === drawableSentinel) drawables.value[2] else right,
            if (bottom === drawableSentinel) drawables.value[3] else bottom
    )
}
```

src/main/java/androidx/widget/TextView.kt
```diff
@@ -0,0 +1,133 @@
...
+    top: Drawable? = compoundDrawables[1],
+    end: Drawable? = compoundDrawables[2],
+    bottom: Drawable? = compoundDrawables[3]
+) {
+    setCompoundDrawables(start, top, end, bottom)
+}
+
+/**
+ * Updates this TextView's Drawables. This version of the method allows using named parameters
+ * to just set one or more Drawables. Use 0 if you do not want a Drawable there.
```
cyrilmottier: Note that this is not consistent with framework's behavior. In particular, passing 0 to a variant of `setCompoundDrawables` will actually throw a `android.content.res.Resources$NotFoundException: Resource ID #0x0` exception.

The question is: Is it OK for android-ktx to enhance/tweak the framework's behavior? In particular, is it OK for `TextView#setCompoundDrawablesWithIntrinsicBounds(0, 0, 0, 0)` to throw an `Exception` but `TextView#updateCompoundDrawablesWithIntrinsicBounds(0, 0, 0, 0)` to execute and clear the compound Drawables?

cyrilmottier: Note that I don't think there's a way to work around this 😢

src/main/java/androidx/widget/TextView.kt
cyrilmottier: BTW: I don't know which one is "lighter"

```kotlin
private val drawableSentinel = DrawableContainer()
```

or

```kotlin
private val drawableSentinel = ColorDrawable(Color.TRANSPARENT)
```

romainguy: I like this solution. @JakeWharton, does this look good to you?

src/main/java/androidx/widget/TextView.kt
romainguy: I think this is where we can safely diverge from the base framework's behavior. We should expand on this in the documentation though to explain exactly what happens when you pass 0.

tomaszrykala: Thank you for your insight @cyrilmottier . This divergence was a conscious decision of mine up for review by the owners of this library if a better solution can be applied. I am open to suggestions.

With regards to the documentation of this method, does it not explain it clearly enough? `Use 0 if you do not want a Drawable there.` . I am not sure what more could we add here @romainguy ?

src/main/java/androidx/widget/TextView.kt
tomaszrykala: Thank you for the suggestion @cyrilmottier . I am happy to apply it if it is required by the maintainers.

tomaszrykala: Tests passing with the proposed solution 👍 
```kotlin
private val drawableSentinel = DrawableContainer()
// ...
fun TextView.updateCompoundDrawablesWithIntrinsicBounds(
    start: Drawable? = compoundDrawables[0],
    top: Drawable? = compoundDrawables[1],
    end: Drawable? = compoundDrawables[2],
    bottom: Drawable? = compoundDrawables[3]
) {
    val drawables = lazy(LazyThreadSafetyMode.NONE) { compoundDrawables }
    setCompoundDrawablesWithIntrinsicBounds(
        if (start === drawableSentinel) drawables.value[0] else start,
        if (top === drawableSentinel) drawables.value[1] else top,
        if (end === drawableSentinel) drawables.value[2] else end,
        if (bottom === drawableSentinel) drawables.value[3] else bottom
    )
}
```

cyrilmottier: :+1:

cyrilmottier: Hmmm I think you forgot to use the `drawableSentinel` in the default parameters.

tomaszrykala: Quite! Corrected 😨

tomaszrykala: Version that can be reused by other `Drawable?` methods?:
```kotlin
private val drawableSentinel = DrawableContainer()
// ...
private fun updateDrawables(
    start: Drawable?, top: Drawable?, end: Drawable?, bottom: Drawable?, srcArray: Array<Drawable>
): Array<Drawable?> {
    val drawables = lazy(LazyThreadSafetyMode.NONE) { srcArray }
    return arrayOfNulls<Drawable>(srcArray.size).apply {
        this[0] = if (start === drawableSentinel) drawables.value[0] else start
        this[1] = if (top === drawableSentinel) drawables.value[1] else top
        this[2] = if (end === drawableSentinel) drawables.value[2] else end
        this[3] = if (bottom === drawableSentinel) drawables.value[3] else bottom
    }
}

// called:

fun TextView.updateCompoundDrawablesWithIntrinsicBounds(
    start: Drawable? = drawableSentinel,
    top: Drawable? = drawableSentinel,
    end: Drawable? = drawableSentinel,
    bottom: Drawable? = drawableSentinel
) {
    val drawables = updateDrawables(start, top, end, bottom, compoundDrawables)
    setCompoundDrawablesWithIntrinsicBounds(drawables[0], drawables[1], drawables[2], drawables[3])
}
// same for updateCompoundDrawablesRelative with compoundDrawablesRelative etc.
```

tomaszrykala: Corrected : ` Using compoundDrawables 4 times to get the default value will possibly involve creating several arrays (one per call to compoundDrawables` < - getting the arrays only once. Sorry I missed that.

src/main/java/androidx/widget/TextView.kt
```diff
@@ -0,0 +1,156 @@
...
+    end: Drawable? = drawableSentinel,
+    bottom: Drawable? = drawableSentinel
+) {
+    val drawables = updateDrawables(start, top, end, bottom, compoundDrawablesRelative)
+    setCompoundDrawablesRelativeWithIntrinsicBounds(
+        drawables[0], drawables[1], drawables[2], drawables[3]
+    )
+}
+
+private fun updateDrawables(
```
cyrilmottier: I understand this is tempting to avoid code duplication. But doing isn't this involving creating a new `Array<Drawable?>` each time is it called? I would duplicate the code in each calling methods or find a way to avoid the array creation.

src/main/java/androidx/widget/TextView.kt
```diff
@@ -0,0 +1,156 @@
...
+    start: Drawable?, top: Drawable?, end: Drawable?, bottom: Drawable?, srcArray: Array<Drawable>
+): Array<Drawable?> {
+    val drawables = lazy(LazyThreadSafetyMode.NONE) { srcArray }
```
cyrilmottier: Note that this is useless here as `srcArray` (and `compoundDrawables`) will already have been called (and the array created). Here again, inlining avoid the issue.

cyrilmottier: Note that you could pass the `Lazy` instead.

src/androidTest/java/androidx/widget/TextViewTest.kt
```diff
@@ -0,0 +1,184 @@
...
+                end = android.R.drawable.btn_default)
+
+        assertNotNull(textView.compoundDrawablesRelative[0])
+        assertNotNull(textView.compoundDrawablesRelative[1])
+        assertNotNull(textView.compoundDrawablesRelative[2])
+        assertNotNull(textView.compoundDrawablesRelative[3])
+    }
+
+    @Test
+    fun updateCompoundDrawablesRelativeWithIntrinsicBoundsWithResourceIds_keepsDrawables() {
```
cyrilmottier: Don't you think adding tests on forbidden values would be useful? Like -2?

src/main/java/androidx/widget/TextView.kt
cyrilmottier: Maybe also mention that -1 will be no-op?

src/main/java/androidx/widget/TextView.kt
tomaszrykala: It does create a new `Array<Drawable?>` per `updateXXXDrawablesXXX()` invocation but given this is a small, 4 index array, I'd personally think that this a good tradeoff for the benefit of losing code repetition. I am open to the suggestion to change it if the maintainers say so.

src/main/java/androidx/widget/TextView.kt
tomaszrykala: Point 1) I need it for the Lazy instance.

Point 2) Indeed, but that would mean that in four call sites we'd have `lazy(NONE) { compoundDrawablesRelative }` or `lazy(NONE) { compoundDrawables}` instead of `compoundDrawablesRelative` and `compoundDrawables` . 

If these changes are required I'm happy to apply them.

src/androidTest/java/androidx/widget/TextViewTest.kt
tomaszrykala: Not a bad idea

src/main/java/androidx/widget/TextView.kt
tomaszrykala: ` Use 0 if you do not want a Drawable there; -1 keeps existing.` ?

romainguy: Maybe something like:

Two special values can be used instead of regular resource IDs:
- 0 to remove an existing drawable (ignored if no drawable exists)
- -1 to keep an existing drawable

Maybe define two constants to make it more readable?

src/main/java/androidx/widget/TextView.kt
romainguy: The allocation doesn't matter much but the runtime matters more than making the code slightly better.

src/main/java/androidx/widget/TextView.kt
```diff
@@ -0,0 +1,160 @@
...
+}
+
+private fun updateDrawables(
+    start: Drawable?,
+    top: Drawable?,
+    end: Drawable?,
+    bottom: Drawable?,
+    srcArray: Array<Drawable>
+): Array<Drawable?> {
+    val drawables = lazy(LazyThreadSafetyMode.NONE) { srcArray }
```
romainguy: I don't understand the need for this lazy() here?

src/androidTest/java/androidx/widget/TextViewTest.kt
tomaszrykala: Thought it could be better to demand correct argument with a `@IntRange` floor value requirement.  Tests added too. Thoughts?
```kotlin
@DrawableRes
const val UPDATE_MODE_REMOVE = 0
@DrawableRes
const val UPDATE_MODE_KEEP = -1
// ...
fun TextView.updateCompoundDrawablesWithIntrinsicBounds(
    @DrawableRes @IntRange(from = UPDATE_MODE_KEEP.toLong()) left: Int = UPDATE_MODE_KEEP,
// ...
```

tomaszrykala:
Rebased on latest master.

oldergod:
What's left to do here?

tomaszrykala:
Can I have an update from the maintainers please if there's anything left to do?

src/androidTest/java/androidx/widget/TextViewTest.kt
TWiStErRob: Are you sure this'll work in all cases?
https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/content/res/ResourceId.java#33

tomaszrykala:
4 months later still nothing. A bit disappointing :(

romainguy:
Sorry Tomasz, nothing left to do at the moment, I just need to find a bit of time to play with the change

DaveMellor:
Hi Tomasz,

Sorry but I'm unclear why I have an email from you. I'm too busy at moment
with a bug in the Intel CPU architecture.

It affects nearly every computer on the planet. See google project zero &
press reports about CPU design flaw.

Regards, Dave

src/main/java/androidx/widget/TextView.kt
```diff
@@ -0,0 +1,172 @@
...
+import android.support.annotation.RequiresApi
+import android.support.v4.content.ContextCompat
+import android.widget.TextView
+
+@DrawableRes
+const val UPDATE_MODE_REMOVE = 0
+@DrawableRes
+const val UPDATE_MODE_KEEP = -1
+
+private val drawableSentinel = DrawableContainer()
```
JakeWharton: Does this turn into a static field in the bytecode? Or do we need to annotate it as `@JvmField`?

src/main/java/androidx/widget/TextView.kt
```diff
@@ -0,0 +1,172 @@
...
+import android.support.annotation.DrawableRes
+import android.support.annotation.IntRange
+import android.support.annotation.RequiresApi
+import android.support.v4.content.ContextCompat
+import android.widget.TextView
+
+@DrawableRes
+const val UPDATE_MODE_REMOVE = 0
+@DrawableRes
+const val UPDATE_MODE_KEEP = -1
```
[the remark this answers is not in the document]
tomaszrykala: Thanks for the reply. By "document" do you mean that instead of 
```
/**
 * - [UPDATE_MODE_REMOVE] to remove an existing drawable (ignored if no drawable exists)
 * - [UPDATE_MODE_KEEP] to keep an existing drawable
```
inline the values to 
```
/**
 * - 0 to remove an existing drawable (ignored if no drawable exists)
 * - -1 to keep an existing drawable
```
?

src/main/java/androidx/widget/TextView.kt
```diff
@@ -0,0 +1,172 @@
...
+ * - [UPDATE_MODE_REMOVE] to remove an existing drawable (ignored if no drawable exists)
+ * - [UPDATE_MODE_KEEP] to keep an existing drawable
+ *
+ * @see TextView.setCompoundDrawablesWithIntrinsicBounds
+ */
+fun TextView.updateCompoundDrawablesWithIntrinsicBounds(
+    @DrawableRes @IntRange(from = UPDATE_MODE_KEEP.toLong()) left: Int = UPDATE_MODE_KEEP,
+    @DrawableRes @IntRange(from = UPDATE_MODE_KEEP.toLong()) top: Int = UPDATE_MODE_KEEP,
+    @DrawableRes @IntRange(from = UPDATE_MODE_KEEP.toLong()) right: Int = UPDATE_MODE_KEEP,
+    @DrawableRes @IntRange(from = UPDATE_MODE_KEEP.toLong()) bottom: Int = UPDATE_MODE_KEEP
```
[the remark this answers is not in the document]
tomaszrykala: So best to skip the range entirely? Since they can be negative as well there's no guarantee what's right or wrong. Will this still be ok for the preset values of 0 (remove) and -1 (keep) though?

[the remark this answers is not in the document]
JakeWharton: 0 is definitely an invalid value and I don't expect -1 to be either as it's reserved to mean "no ID".

src/main/java/androidx/widget/TextView.kt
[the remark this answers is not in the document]
JakeWharton: Precisely. Although ideally not as the very first thing in the KDoc of the method but more the very last, or as part of the `@param`.

src/main/java/androidx/widget/TextView.kt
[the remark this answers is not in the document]
tomaszrykala: I've removed the range requirement and so any value will go. The default argument is -1 (remove) as it's documented in the KDoc. If we want the value that represents "remove" changed let's agree on one. `Int.MIN_VALUE` etc?

src/main/java/androidx/widget/TextView.kt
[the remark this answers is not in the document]
tomaszrykala: Done.

PR #161, opened 2018-02-06
Add FragmentManager extension for transactions

rom4ek:
Add extension to execute `FragmentTransaction` in a handy way
 ```
 fragmentManager.transact {
    remove(fragmentOne)
    add(R.id.fragment_container, fragmentTwo)
 }
```


rom4ek:
I signed it!

JakeWharton:
Thanks for taking the time to make the PR (with tests!), but we would like to encourage that developers only use the support library fragments. The next version of Android will deprecate the version of fragments that are part of the platform. Thus, we aren't going to add any extensions to support them in this project.

PR #165, opened 2018-02-06
Add SeekBar listener extensions

bernaferrari:
Two things to notice:
- I was only able to make the test for onProgressChanged, since onStartTrackingTouch and onStopTrackingTouch deal directly with UI and private methods.
- I'm not sure the name from a method is the best one. The original is "onStartTrackingTouch", I made it become "onStartTracking", is this ok?

### DIFF
--- a/api/0.1.txt
+++ b/api/0.1.txt
@@ -671,3 +671,13 @@ package androidx.view {
 
 }
 
+package androidx.widget {
+
+  public final class SeekBarKt {
+    ctor public SeekBarKt();
+    method public static final android.widget.SeekBar doOnStartTracking(android.widget.SeekBar, kotlin.jvm.functions.Function1<? super android.widget.SeekBar,kotlin.Unit> action);
+    method public static final android.widget.SeekBar doOnStopTracking(android.widget.SeekBar, kotlin.jvm.functions.Function1<? super android.widget.SeekBar,kotlin.Unit> action);
+    method public static final android.widget.SeekBar doOnChanged(android.widget.SeekBar, kotlin.jvm.functions.Function2<? super android.widget.SeekBar,? super Integer,? super Boolean,kotlin.Unit> action);
+  }
+
+}
--- a/src/androidTest/java/androidx/widget/SeekBarTest.kt
+++ b/src/androidTest/java/androidx/widget/SeekBarTest.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.widget
+
+import android.support.test.InstrumentationRegistry
+import android.widget.SeekBar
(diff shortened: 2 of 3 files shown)


bernaferrari:
I signed it!

src/main/java/androidx/widget/SeekBar.kt
```diff
@@ -0,0 +1,71 @@
...
+ * @return the [SeekBar.OnSeekBarChangeListener] added to the SeekBar
+ * @see SeekBar.OnSeekBarChangeListener.onProgressChanged
+ */
+fun SeekBar.doOnChanged(action: (seekBar: SeekBar, progress: Int, fromUser: Boolean) -> Unit): SeekBar.OnSeekBarChangeListener =
+    OnSeekBarChangeListener(onChanged = action)
+
+/**
+ * Set the listener to this SeekBar using the provided actions.
+ */
+fun SeekBar.OnSeekBarChangeListener(
```
romainguy: A method name must start with a lower case. addListener() would be a better name (or addChangeListener).

src/main/java/androidx/widget/SeekBar.kt
```diff
@@ -0,0 +1,71 @@
...
+fun SeekBar.doOnStopTracking(action: (seekBar: SeekBar) -> Unit): SeekBar.OnSeekBarChangeListener =
+    OnSeekBarChangeListener(onStopTracking = action)
+
+/**
+ * Add an action which will be invoked when the SeekBar has the progress changed.
+ *
+ * @return the [SeekBar.OnSeekBarChangeListener] added to the SeekBar
+ * @see SeekBar.OnSeekBarChangeListener.onProgressChanged
+ */
+fun SeekBar.doOnChanged(action: (seekBar: SeekBar, progress: Int, fromUser: Boolean) -> Unit): SeekBar.OnSeekBarChangeListener =
```
romainguy: So the way these extensions work, calling doOnStartTracking then doOnStopTracking will cause the first action (onStartTracking) to be dropped

bernaferrari: Oh, crap. Any ideas on how to improve this?

src/main/java/androidx/widget/SeekBar.kt
bernaferrari: I agree (there is even a lint warning). I tried to follow the Animator (and others) convention, but SeekBar does not have an internal lowercased method.. It only allows to set the value, the other methods add to an ArrayList.

romainguy: You can call it `setOnSeekBarChangeListener` since it has a different signature.

bernaferrari: Fixed on latest commit (f133743). Thanks!

api/0.1.txt
```diff
@@ -671,3 +671,13 @@ package androidx.view {
 
 }
 
+package androidx.widget {
```
romainguy: You should not be changing this file I believe.

src/androidTest/java/androidx/widget/SeekBarTest.kt
```diff
@@ -0,0 +1,38 @@
...
+
+class SeekBarTest {
+    private val context = InstrumentationRegistry.getContext()
+    private val view = SeekBar(context)
+
+    @Test
+    fun testDoOnChanged() {
+        var called = false
+        view.progress = 0
+        view.doOnChanged { _, _, _ ->
```
romainguy: Please add tests for all the new APIs

api/0.1.txt
bernaferrari: Ops! I saw a PR yesterday modifying it, but it also reverted later. Reverted.

src/androidTest/java/androidx/widget/SeekBarTest.kt
bernaferrari: Added, sorry for the inconvenience.

Issue #138, opened 2018-02-05
Add an extension property to get the children of a ViewGroup as a Iterable


quiro91:
```kotlin
val ViewGroup.children: Iterable<View>
    get() = (0 until childCount).map { getChildAt(it) }
```

This would enable to use all the operators from the Kotlin standard library like filter and so on instead of having to rewrite them all as ViewGroup extension functions

JakeWharton:
Yeah we used to have something like this until we switched to `operator fun iterator()`. It's a shame there isn't a structural typing such that all the existing collection extensions could apply. Is there any reason to expose a `List` rather than an `Iterable`?

quiro91:
Honestly no, I agree that `Iterable` seems more appropriate. Will change the title of the issue

kevinmost:
Any thoughts on it being a `Sequence` instead? This is what we implemented in our project, I'm not sure if the laziness of a sequence is more or less desirable though.

```kotlin
/**
 * Returns a [Sequence] that iterates through the children of this view
 */
val ViewGroup.children: Sequence<View>
    get() {
        var currentIndex = -1
        return generateSequence {
            currentIndex++
            getChildAt(currentIndex)
        }
    }
```

mokkun:
IMHO it should always return a `Iterable`. If you need it as a sequence you could always do `myView.children.asSequence()`.

JakeWharton:
The opposite of that is also true in asIterable()

JakeWharton:
Yeah. Plus there actually are a few eager operators on Sequence like `any`
which are commonly used.

JakeWharton:
Was merged.

PR #168, opened 2018-02-06
Add children val to ViewGroup which returns a Sequence<View>

panpanini:
References #138

Adds a `children` val extension to ViewGroup which returns a Sequence of child views

I'm not too sure about the naming of the test method, so if someone has a better suggestion :pray:

### DIFF
--- a/src/androidTest/java/androidx/view/ViewGroupTest.kt
+++ b/src/androidTest/java/androidx/view/ViewGroupTest.kt
@@ -203,11 +203,8 @@ class ViewGroupTest {
         val views = listOf(View(context), View(context), View(context))
         views.forEach { viewGroup.addView(it) }
 
-        var index = 0
-        viewGroup
-                .children
-                .forEach { child ->
-                    assertSame(views[index++], child)
+        viewGroup.children.forEachIndexed { index, child ->
+                    assertSame(views[index], child)
                 }
     }
 
--- a/src/androidTest/java/androidx/view/ViewGroupTest.kt
+++ b/src/androidTest/java/androidx/view/ViewGroupTest.kt
@@ -204,8 +204,8 @@ class ViewGroupTest {
         views.forEach { viewGroup.addView(it) }
 
         viewGroup.children.forEachIndexed { index, child ->
-                    assertSame(views[index], child)
-                }
+            assertSame(views[index], child)
+        }
     }
 
     @Test fun setMargins() {
--- a/src/androidTest/java/androidx/view/ViewGroupTest.kt
+++ b/src/androidTest/java/androidx/view/ViewGroupTest.kt
@@ -199,6 +199,18 @@ class ViewGroupTest {
         }
     }
 
+    @Test fun children() {
+        val views = listOf(View(context), View(context), View(context))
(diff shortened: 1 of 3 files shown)


src/androidTest/java/androidx/view/ViewGroupTest.kt
```diff
@@ -199,6 +199,18 @@ class ViewGroupTest {
         }
     }
 
+    @Test fun children() {
+        val views = listOf(View(context), View(context), View(context))
+        views.forEach { viewGroup.addView(it) }
+
+        var index = 0
```
JakeWharton: use `forEachIndexed`

src/androidTest/java/androidx/view/ViewGroupTest.kt
```diff
@@ -199,6 +199,18 @@ class ViewGroupTest {
...
+        viewGroup
+                .children
+                .forEach { child ->
```
JakeWharton: You can put all these on the same line as `viewGroup`

src/main/java/androidx/view/ViewGroup.kt
```diff
@@ -71,6 +71,16 @@ operator fun ViewGroup.iterator() = object : MutableIterator<View> {
     override fun remove() = removeViewAt(--index)
 }
 
+/** Returns a [Sequence] over the child views in this view group. */
+val ViewGroup.children: Sequence<View>
+    get() {
+        var index = -1
+        return generateSequence {
```
JakeWharton: This has quite a bit of overhead. Why don't we just return an `object : Sequence` whose function delegates to our existing `iterator()` method instead.

panpanini: would this be the same as something like:
``` kotlin
val ViewGroup.children: Sequence<View>
  get() = iterator().asSequence()
```

JakeWharton: I'd need to check what `asSequence` does. It probably prevents multiple iteration which my proposed solution doesn't.

panpanini: yeah it does, thats expected for a Sequence right?

JakeWharton: It's not required, no.

src/androidTest/java/androidx/view/ViewGroupTest.kt
```diff
@@ -199,6 +199,15 @@ class ViewGroupTest {
         }
     }
 
+    @Test fun children() {
+        val views = listOf(View(context), View(context), View(context))
+        views.forEach { viewGroup.addView(it) }
+
+        viewGroup.children.forEachIndexed { index, child ->
+                    assertSame(views[index], child)
```
romainguy: Your indentation is incorrect FYI

JakeWharton:
You still need to run the `updateApi` task and commit its changes.

panpanini:
ah whoops, I ran the task but I'm not sure if I committed the results. I'll check that now

JakeWharton:
Squashed and merged. Thanks!

PR #177, opened 2018-02-06
Use generics to get rid of casting

bina1204:
It is better to use generics like below.

``` 
imageView.doOnPreDraw {
    (it as ImageView).doSomething()
}
```


JakeWharton:
Please retain an overload with the old signature and a deprecated annotation with a visibility of hidden to maintain binary compatibility as well.

davethomas11:
Does it make sense to upgrade signature for ```doOnLayout``` and ```doOnNextLayout``` to use generics as well?

JakeWharton:
Sure!

bina1204:
@romainguy Thank you for your review!
I updated `current.txt` and added some tests.

PR #180, opened 2018-02-06

scrobot:
For any cases to determine view visibility state, such as 
``
view.visibility { 
    someAction()
}
``


arnis71:
@scrobot this PR looks better [#146](https://github.com/android/android-ktx/pull/146). Using lambdas for view visibility is an overkill. Plus no way to toggle the Gone state.

PR #182, opened 2018-02-06

scrobot:
For any cases to determine view visibility state, such as
view.visibility { someAction() }


.idea/codeStyleSettings.xml
```diff
@@ -1,71 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
romainguy: Don't modify this file

.idea/copyright/AOSP.xml
```diff
@@ -1,6 +0,0 @@
-<component name="CopyrightManager">
```
romainguy: Don't modify

.idea/copyright/profiles_settings.xml
```diff
@@ -1,3 +0,0 @@
-<component name="CopyrightManager">
```
romainguy: Don't modify

src/main/java/androidx/view/View.kt
```diff
@@ -172,3 +172,23 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
     }
     return Bitmap.createBitmap(width, height, config).applyCanvas(::draw)
 }
+
+/**
+ * Sets the view's visibility by Boolean lambda. This version of the method sets all axes to the provided size.
```
romainguy: The second part of the doc does not match this function ("This version…")

src/main/java/androidx/view/View.kt
```diff
@@ -172,3 +172,23 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+ *
+ * @see View.setVisibility
+ */
+@RequiresApi(14)
```
romainguy: Why does it require API 14?

src/main/java/androidx/view/View.kt
```diff
@@ -172,3 +172,23 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+inline fun View.visibility(expression: () -> Boolean?) {
```
romainguy: It is odd (and undocumented) that `false` would lead to the view being GONE

src/main/java/androidx/view/View.kt
```diff
@@ -172,3 +172,23 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+    this.visibility = if(expression.invoke() == true) View.VISIBLE else View.GONE
```
romainguy: Why .invoke() and not just expression()?

src/main/java/androidx/view/View.kt
romainguy: Why is the boolean nullable?

src/main/java/androidx/view/View.kt
romainguy: No need for == true

romainguy:
You also need to run `updateApi`

src/main/java/androidx/view/View.kt
scrobot: sory, it was the test to fix problem with gradle check) forgot to remove)

PR #189, opened 2018-02-06
Extensions for startActivity methods

MartinDevi:
Adds fluency to be able to start new activities using statements like
```
startActivity<MyActivity>()
startActivity<MyActivity>() {
    putExtra("key", "value")
}
startActivity<MyActivity>(ActivityOptions.makeBasic().toBundle())
```

### DIFF
--- a/src/main/java/androidx/app/Activity.kt
+++ b/src/main/java/androidx/app/Activity.kt
@@ -1,0 +1,19 @@
+package androidx.app
+
+import android.app.Activity
+import android.content.Intent
+import android.os.Bundle
+import android.support.annotation.RequiresApi
+
+/**
+ * Syntactic sugar for [Activity.startActivityForResult(Intent, Int)][Activity.startActivityForResult].
+ */
+inline fun <reified T> Activity.startActivityForResult(code: Int, block: Intent.() -> Unit = {}) =
+        startActivityForResult(Intent(this, T::class.java).apply(block), code)
+
+/**
+ * Syntactic sugar for Activity.startActivityForResult(Intent, Int, Bundle)[Activity.startActivityForResult].
+ */
+@RequiresApi(16)
+inline fun <reified T> Activity.startActivityForResult(code: Int, options: Bundle, block: Intent.() -> Unit = {}) =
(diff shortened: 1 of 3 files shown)


src/main/java/androidx/app/Activity.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import android.app.Activity
+import android.content.Intent
+import android.os.Bundle
+import android.support.annotation.RequiresApi
+
+/**
+ * Syntactic sugar for [Activity.startActivityForResult(Intent, Int)][Activity.startActivityForResult].
+ */
+inline fun <reified T> Activity.startActivityForResult(code: Int, block: Intent.() -> Unit = {}) =
```
tetedoie: It's definitely the first extension that i wanted to add. Actually your reified parameter must be an activity so it should be <reified T : Activity>

tetedoie: Same for all other start extensions. In addition, we must handle startActivityForResult for fragments as well. I'm not sure if the support library is in the scope of this library.

MartinDevi: Agree on both points. I added the upper bound and the extensions for regular fragments. From what I can tell the support library isn't in the scope right now.

sourabhv:
An extension to create `Intent` would be more useful.

JakeWharton:
I agree. It would then apply to things like services, broadcasts, notifications, etc. We don't need to make starting things easier. That's already easy. We need to make creating intents easier. Plus there's a million start* methods and I don't want to have to track them all.

joshfriend:
> Plus there's a million start* methods and I don't want to have to track them all.

This is the problem I always have with Anko's [numerous](https://github.com/Kotlin/anko/blob/93fd14ea606f861082b364df43c76221b135552f/anko/library/static/commons/src/Intents.kt#L30-L43) [`startActivity*` methods](https://github.com/Kotlin/anko/blob/ed9c89105c0361f418fb49046a59303ba0257852/anko/library/static/supportV4/src/SupportIntents.kt#L38-L44). It's always difficult to import the right one

PR #195, opened 2018-02-06
Add postApply/postDelayedApply for View

dewarder:
Add versions of post/postDelayed which accept View as receiver in block. It is helpful when you want to apply some changes from another thread:

```kotlin
val refreshLayout: SwipeRefreshLayout = ...
refreshLayout.postApply {
    isRefreshing = false
    scaleX = 0f
    scaleY = 0f
}
```


dewarder:
I've also added tests for `post`/`postDelayed` extensions and placed them all into separate file due to they need visible window.

PR #198, opened 2018-02-06
views visibility and invisibility ext. functions

scrobot:


JakeWharton:
If we have #146 you can put an expression or function invocation on the right hand side of an assignment to one of the setters. Let's focus discussion there.

Issue #143, opened 2018-02-05
`Toast` extensions on `Context`


n8ebel:
I've seemed to add this to each Kotlin project I've worked on

`fun Context.shortToast(msg: CharSequence) = Toast.makeText(this, msg, Toast.LENGTH_SHORT).show()`

`fun Context.longToast(msg: CharSequence) = Toast.makeText(this, msg, Toast.LENGTH_LONG).show()`

cleans up the api a bit and avoids the need to remember to call `show()`

egor-n:
This may be handy, but apps should really stop using `Toast` since they don't show up if notifications are disabled for an app.

bernaferrari:
Egor, I'm curious, what is the recommended way? Snackbar?

MGaetan89:
And instead of two methods, why not one with a `duration` parameter?

abbath0767:
i thing good solution for duration is default fun argument, like this:
```
fun Context.toast(messageId: Int, duration: Int = Toast.LENGTH_LONG) =
        Toast.makeText(this, messageId, duration).show()
```

n8ebel:
a single function with a duration arg & a default value is a nice idea 👍    whether it's added here or not, I'll have to incorporate that 😄

jshvarts:
I normally something similar to what @abbath0767 suggested. 
    fun Context.toast(@StringRes resId: Int, duration: Int = Toast.LENGTH_LONG) =
        Toast.makeText(this, resId, duration).show()

JakeWharton:
> avoids the need to remember to call `show()`

Lint guards against this.

frederickcxa:
What about this?

```
fun Context.toast(content: Any, length: Int = Toast.LENGTH_SHORT): Toast {
    val message = when (content) {
        is String -> content
        is Int -> getString(content)
        else -> throw IllegalArgumentException("Content must be a String or a String resource id.")
    }

    return Toast.makeText(this, message, length).also { it.show() }
}
```

Here we can pass a `String`|`String resource id` to the method, also we return the toast object back to the caller. 😃

JakeWharton:
That seems like it sacrifices type safety for no benefit whatsoever. Why wouldn't you use an overload?

frederickcxa:
You right, type safety for the win.

frederickcxa:
We could use `also` to be able to show and return the `Toast`.

```
fun Context.toast(@StringRes resId: Int, duration: Int = Toast.LENGTH_LONG) =
[...]

PR #199, opened 2018-02-06
Added extension functions for long and short toasts.

tobiasschuerg:


MGaetan89:
There is a discussion running in #143 about adding `Toast` support.

tobiasschuerg:
oh, didn't see that

PR #201, opened 2018-02-06
Add readList()/readSet()/read()/readFirst() cursor extensions

linsalion:
Simplify reading data from cursor

```
val uniqueNames = cursor.readSet { it.getString("name") }

val books = cursor.readList { 
    Book(
        it.getLong("id") 
        it.getString("title") 
        it.getString("author") 
    }    
}

val richestPersonName = cursor.readFirst { it.getString("name") }
```

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -81,7 +81,7 @@ package androidx.database {
     method public static final String getString(android.database.Cursor, String columnName);
     method public static final String? getStringOrNull(android.database.Cursor, int index);
     method public static final error.NonExistentClass getStringOrNull(android.database.Cursor, String columnName);
-    method public static final <R> java.util.Collection<R> read(android.database.Cursor cursor, java.util.Collection<R> outCollection, kotlin.jvm.functions.Function1<? super android.database.Cursor,? extends R> itemReadFunc);
+    method public static final <R> java.util.Collection<R> read(android.database.Cursor, java.util.Collection<R> outCollection, kotlin.jvm.functions.Function1<? super android.database.Cursor,? extends R> itemReadFunc);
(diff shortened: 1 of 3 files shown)


JakeWharton:
The reason these weren't included is that it encourages the use of `Cursor` for larger tasks where something like Room or SQL Brite/Delight can do a much better job at mapping lists or individual items. Can you describe your use case?

linsalion:
Other database frameworks help to avoid working with `Cursor` at all and use models instead.
But if there is a need to use `Cursor` or there is no possibility to use Room or SQL Brite/Delight,
then reading from `Cursor` produces a lot of boilerplace code like this:
```
cursor.use {
    cursor.moveToFirst()

    while (!cursor.isAfterLast) {
        val item = ... 
        ...
        reading from cursor
        ...
        outCollection.add(item)
        cursor.moveToNext()
    }
}
```

Almost 100% of the time you need to get a populated list or just one element (when the query contains "limit 1").
These extensions really simplify it. I have also added examples in pull request description.

davethomas11:
@JakeWharton  Hey Jake, I was working with DownloadManager today and was thinking about the sentiment with Cursor extension functions. [DownloadManager.query](https://developer.android.com/reference/android/app/DownloadManager.html#query(android.app.DownloadManager.Query)) returns a Cursor, some extension functions would be useful in this case. Also, I can remember doing media queries (checking for videos and images) before, and a Cursor was involved.

I generally agree with your opinion when it comes to using other solutions for database code, and avoiding Curors. Food for thought though, there are other APIs that return Cursors. 

Unless DownloadManager, and Media queries are things of the past and there are different ways of doing those kinds of tasks?

davethomas11:
Disregard my previous statements :facepalm:. What is currently present in the Cursor extension provided is sufficient, and works great =).

JakeWharton:
Going to close based on my previous comment for now.

PR #215, opened 2018-02-06
Replace `for` with `foreach` in ViewGroup

rkowase:
Replace `for` with `foreach` in ViewGroup.


JakeWharton:
This generates significantly less suitable bytecode as it allocates a `Range` and then uses an `Iterator` to traverse it.

PR #217, opened 2018-02-06
shared preference api update

UmangChamaria:
The current API says edit which is ambiguous, by the name it is not evident whether it is `apply()` or `commit()`.
Also  currently there is no support for commit incase an app wants to use it.
Changes :  
1. API name changed for `edit()` -> `apply()`
2. Added support for `commit()`


JakeWharton:
If we're going to change this, and I'm not convinced we should, you'll need to retain the old method and test with a deprecation and replace-with annotation to migrate users to the new one.

romainguy:
Isn't this ambiguous/confusing because of Kotlin's apply{}?

UmangChamaria:
@JakeWharton Updated the PR, failing test cases are for other APIs not related to the changes in PR
@romainguy I agree it can be confusing with apply{} but so is `edit()` In my opinion this is more self-explanatory (might be wrong)

DSteve595:
What about `applyEdits { }` and `commitEdits { }`?

simonschiller:
I guess that in the majority of the cases `apply` would be used. I would keep `edit { }` and add another function for the few cases where you need to use `commit`.
Maybe something like `editSync { }` or `editWithResult { }`?

JakeWharton:
How about `edit(commit: Boolean = false, block: Editor.() -> Unit)`?

UmangChamaria:
@JakeWharton Is this PR worth a merge or should I close it?
Note: The failing test case is no where related to the code change in the PR. That was failing even without the change.

src/main/java/androidx/content/SharedPreferences.kt
```diff
@@ -20,16 +20,27 @@ import android.content.SharedPreferences
...
+ * Default behaviour is [SharedPreferences.Editor.apply]
  *
+ * To [SharedPreferences.Editor.apply] changes
  * ```
  * prefs.edit {
  *     putString("key", value)
  * }
  * ```
+ * To [SharedPreferences.Editor.commit] changes
+ * prefs.edit {
```
JakeWharton: This should be `edit(commit = true) {`

src/main/java/androidx/content/SharedPreferences.kt
```diff
@@ -20,16 +20,27 @@ import android.content.SharedPreferences
...
+ *     putString("key", value)
+ * }
  */
-inline fun SharedPreferences.edit(action: SharedPreferences.Editor.() -> Unit) {
+inline fun SharedPreferences.edit(
+    commit: Boolean = false,
+    action: SharedPreferences.Editor.() -> Unit) {
```
JakeWharton: nit: `) {` should be on its own line

src/main/java/androidx/content/SharedPreferences.kt
```diff
@@ -20,16 +20,27 @@ import android.content.SharedPreferences
...
     val editor = edit()
     action(editor)
-    editor.apply()
+    when (commit) {
```
JakeWharton: Use an `if`/`else` here. I don't think the compiler can eliminate `when`s whereas it definitely can for `if`/`else`.

src/main/java/androidx/content/SharedPreferences.kt
```diff
@@ -20,16 +20,27 @@ import android.content.SharedPreferences
 
 /**
  * Allows editing of this preference instance with a call to
- * [SharedPreferences.Editor.apply] to persist the changes.
+ * [SharedPreferences.Editor.apply] or [SharedPreferences.Editor.commit] to persist the changes.
+ * Default behaviour is [SharedPreferences.Editor.apply]
```
JakeWharton: nit: missing period

src/main/java/androidx/content/SharedPreferences.kt
JakeWharton: missing `` ``` `` before and after code snippet

src/androidTest/java/androidx/content/SharedPreferencesTest.kt
```diff
@@ -34,4 +34,15 @@ class SharedPreferencesTest {
         assertEquals("test_value", preferences.getString("test_key1", null))
         assertEquals(100, preferences.getInt("test_key2", 0))
     }
+
+    @Test fun editAndApply() {
```
JakeWharton: should be called `editAndCommit`

UmangChamaria:
@JakeWharton Updated the pull request based on the comments.
The failing test is not from the PR.
Thanks

JakeWharton:
I'll look tomorrow. Thanks!

JakeWharton:
Squashed and merged manually.

PR #218, opened 2018-02-06
Add private modifier to fix lint warning

rkowase:
Add private modifier in `AtomicFileTest.kt` to fix lint warning.


JakeWharton:
JUnit requires this to be public.

PR #219, opened 2018-02-06
Add annotation to suppress MemberVisibilityCanPrivate warning

rkowase:
Add annotation to suppress MemberVisibilityCanPrivate warning because JUnit requires this to be public.

ref. #218


dewarder:
I think, it will be better to annotate property with `@JvmField` as it is done in `TransitionTest`. Such changes will also hide lint warning.

rkowase:
Thank you for your suggestion. I'll send another PR, then I close this PR

PR #222, opened 2018-02-06
add File#toUri()

mightyfrog:
Another frequent use case of Uri conversion, like String#toUri()


JakeWharton:
You need to run `updateApi` task and commit the result.

mightyfrog:
api/current.txt?

mightyfrog:
api/current.txt added

JakeWharton:
Thanks!

PR #225, opened 2018-02-06
Add show/hide extension methods to View for concise alpha animations

AJScanlan:
Animating alpha values is a thing developers will do hundreds of times and while the ViewPropertyAnimator is great and easy to use it would be nice to just be able to `view.show()` to animate to 1.0f alpha.

Since 1.0f and 0.0f are by far the most common alpha values to animate to I believe this would be a handy extension method.

If I missed something or my naming conventions are off please let me know, I would be happy to fix it.

For testing I wanted to add a test for the `onFinished` functionality but articles online told me async tests are controversial so I've erred on the side of simple tests.


AJScanlan:
I signed it!

JakeWharton:
I don't think these add any real value and they bake in defaults that are not generally applicable. You can already do `view.animate().alpha(0)` which provides the ability to customize the duration, interpolator, delay, and listeners via other extensions in this project.

PR #228, opened 2018-02-06
Bug with SparseArray

SavinMike:
If `SparseArray` (or it modification such as `SparseLongArray`) contains an element, then `indexOfKey` function returns `-2`. So the methods `contains` and `containsKey` worked incorrectly

### DIFF
--- a/src/androidTest/java/androidx/util/LongSparseArrayTest.kt
+++ b/src/androidTest/java/androidx/util/LongSparseArrayTest.kt
@@ -42,6 +42,16 @@ class LongSparseArrayTest {
         assertTrue(1L in array)
     }
 
+    @Test fun containsOperatorWithValue() {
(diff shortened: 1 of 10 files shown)


JakeWharton:
Nice find.

SavinMike:
@JakeWharton Also you have two test files for LongSparseArray [first](https://github.com/android/android-ktx/blob/master/src/androidTest/java/androidx/util/LongSparseArrayTest.kt) and [second](https://github.com/android/android-ktx/blob/master/src/androidTest/java/androidx/util/SparseLongArrayTest.kt)

JakeWharton:
Those are different types!

SavinMike:
@JakeWharton Updated for LongSparseArray

JakeWharton:
Squashed and merged. Thanks!

PR #230, opened 2018-02-06
File.toUri() to API 24+

jeziellago:
Uri.fromFile() throw FileUriExposedException on API 24+


JakeWharton:
`Uri.fromFile` does not throw an exception on API 24+. If you use a `Uri` to represent a `File` through a content resolver you'll get an exception, but that's completely orthogonal. Representing a file as a URI is perfectly valid otherwise and some libraries actually rely on it.

We don't have access to `FileProvider` in this library so we cannot use this mechanism.

PR #232, opened 2018-02-06
added Menu and its test

hendraanggrian:
Following convention of `ViewGroup.kt`, `Menu.kt` does something similar to `MenuItem` as `ViewGroup` does to `View.`


hendraanggrian:
I signed it!

api/0.1.txt
```diff
@@ -639,6 +639,17 @@ package androidx.util {
 
 package androidx.view {
 
+  public final class MenuKt {
```
JakeWharton: Do not change this file.

src/main/java/androidx/view/Menu.kt
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Returns the menu at `index`.
+ *
+ * @throws IndexOutOfBoundsException if index is less than 0 or greater than or equal to the count.
+ */
+inline operator fun Menu.get(index: Int): MenuItem = getItem(index)
+
+/** Returns `true` if `item` is found in this menu. */
+operator fun Menu.contains(item: MenuItem): Boolean {
+    return (0 until size()).any { getItem(it) == item }
```
JakeWharton: Use a normal loop here. This creates a Range and two Iterators which are unnecessary.

src/androidTest/java/androidx/view/MenuTest.kt
```diff
@@ -0,0 +1,107 @@
...
+import android.support.test.InstrumentationRegistry
+import android.view.MenuItem
+import android.widget.Toolbar
+import androidx.assertThrows
+import androidx.fail
+import com.google.common.truth.Truth.assertThat
+import junit.framework.Assert.assertEquals
+import junit.framework.Assert.assertFalse
+import junit.framework.Assert.assertNotNull
+import junit.framework.Assert.assertTrue
```
JakeWharton: Replace all these with `org.junit` imports.

src/androidTest/java/androidx/view/MenuTest.kt
```diff
@@ -0,0 +1,107 @@
...
+import org.junit.Assert.assertSame
+import org.junit.Test
+
+class MenuTest {
+    private val menu = Toolbar(InstrumentationRegistry.getContext()).menu
+
+    @Test fun get() {
+        val item = menu.add("")
+        assertNotNull(menu[0])
+        assertEquals(item, menu[0])
```
JakeWharton: Both of these can be captured in a single `assertSame` call

src/androidTest/java/androidx/view/MenuTest.kt
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Test fun isEmpty() {
+        assertTrue(menu.isEmpty())
+        menu.add("")
+        assertFalse(menu.isEmpty())
+    }
+
+    @Test fun isNotEmpty() {
+        menu.add("")
```
JakeWharton: assertFalse before this call for parity with isEmpty test

src/androidTest/java/androidx/view/MenuTest.kt
```diff
@@ -0,0 +1,107 @@
...
+
+        val items = mutableListOf<MenuItem>()
+        menu.forEachIndexed { index, item ->
+            assertEquals(index, items.size)
+            items += item
+        }
+        assertThat(items).containsExactly(item1, item2)
+    }
+
+    @Test fun iterator() {
```
JakeWharton: Nothing tests remove behavior of the iterator

hendraanggrian: I'm sorry I don't quite get this one, everything else is fixed as suggested.

src/main/java/androidx/view/Menu.kt
```diff
@@ -30,7 +30,12 @@ inline operator fun Menu.get(index: Int): MenuItem = getItem(index)
 
 /** Returns `true` if `item` is found in this menu. */
 operator fun Menu.contains(item: MenuItem): Boolean {
-    return (0 until size()).any { getItem(it) == item }
+    for (index in 0 until size()) {
+        if (getItem(index) == item) {
+            return false
```
rom4ek: Should be `return true`

hendraanggrian: my bad, just fixed it

api/0.1.txt
```diff
@@ -670,4 +670,3 @@ package androidx.view {
   }
 
 }
-
```
JakeWharton: revert changes to this file please

src/androidTest/java/androidx/view/MenuTest.kt
```diff
@@ -0,0 +1,146 @@
...
+
+    @Test fun get() {
+        val item = menu.add("")
+        assertSame(item, menu[0])
+    }
+
+    @Test fun contains() {
+        val item1 = menu.add("")
+        assertTrue(item1 in menu)
+        assertFalse(item1 !in menu)
```
JakeWharton: This second assertion is testing kotlin and can be removed

src/androidTest/java/androidx/view/MenuTest.kt
```diff
@@ -0,0 +1,146 @@
...
+
+        val item2 = menu.add("")
+        assertTrue(item2 in menu)
+        assertFalse(item2 !in menu)
```
JakeWharton: ditto

src/androidTest/java/androidx/view/MenuTest.kt
```diff
@@ -0,0 +1,146 @@
...
+        assertFalse(item1 in menu)
+        assertEquals(1, menu.size())
+
+        assertSame(item2, iterator.next())
+        iterator.remove()
+        assertFalse(item2 in menu)
+        assertEquals(0, menu.size())
+    }
+
+    @Test fun iteratorForEach() {
```
JakeWharton: This is testing the same thing as `iterator()` but with a syntax. The test can be removed.

hendraanggrian: Removed.

api/0.1.txt
hendraanggrian: Reverted.

src/androidTest/java/androidx/view/MenuTest.kt
hendraanggrian: Removed. It was originally there because I followed `ViewGroupTest.contains()`.

JakeWharton:
Squashed, fixed up a few things, and merged. Thanks!

Issue #132, opened 2018-02-05
DP utitilies on Int and Float


imminent:
Calculating dimensions, such as, SP and DP require the `DisplayMetrics`, so you can't do something too simple without - say - adding Extensions inside of `Activity` and such directly. But you can still make some progress similar to the `Duration` extensions doing something like this:
```kotlin
/**
 * Applies density dimensions to the [Int]
 */
fun Int.dp(displayMetrics: DisplayMetrics) = toFloat().dp(displayMetrics).toInt()

/**
 * Applies density dimensions to the [Float]
 */
fun Float.dp(displayMetrics: DisplayMetrics) = TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP, this, displayMetrics)
```

Which would be called like `9.dp(displayMetrics)`.

JakeWharton:
I think we would call these `toDp` as conversion factory extensions are prefixed with "to".

imminent:
Ah, that's fine, updating the description to `to`. 👍

nohitme:
Just curious, why is it applied on Int & Float instead of on DisplayMetrics ike `displayMetrics.toDp(9)`?

That way you can also name the param "pixel" so it's less confusing.

daberni:
I am not sure if it should be treated/interpreted as a conversion. mostly you simply want do pass a value in dp to a method which takes pixels. so `9.toDp()` I would assume converting `9px` to `Ndp` but what we want is to be get the value `9` get treated as `dp` and therefore I would prefer `9.asDp()`

jacklt:
I wrote this utility:

```
fun Context.toPixelFromDip(value: Float) = TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP, value, resources.displayMetrics)
```

```
fun View.toPixelFromDip(value: Float) = context.toPixelFromDip(value)
```

Pro:
- `toPixelFromDip()` is very clear the origin and the destination
- It's handy when you call this method from a `Context` or (more common) a `View`, ex: `View().apply { translationX = toPixelFromDip(2) }`

lupajz:
Wouldn't accessing system `Resources` just be sufficiet? So we can have something like:
```
val Int.dp: Int // or just return Float here
    get() = (this * Resources.getSystem().displayMetrics.density).toInt()
```
and use it in place like `val side = 42.dp` ?

jacklt:
[...]

PR #233, opened 2018-02-06
Add Int.dp.

chris-horner:
An extension property that converts an integer of dp into px, allowing for calls such as
```kotlin
view.updatePadding(top = 16.dp)
```
Few points that are probably worth discussing:
- Are there any issues with statically accessing `Resources` with `Resources.getSystem()`?
- Should this be a property or a method?
- Should this return an `Int` or a `Float`? Android's APIs sometimes require one or the other. I've personally found `Int` to be more useful during day-to-day dev.

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -36,6 +36,11 @@ package androidx.content {
 
 package androidx.content.res {
 
+  public final class ResourcesKt {
+    ctor public ResourcesKt();
+    method public static final int getDp(int);
+  }
+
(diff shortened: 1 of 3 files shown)


JakeWharton:
There are discussions about this in #132 which cover why `Resources.system()` isn't suitable for this.

chris-horner:
Gotcha. Sounds like people are already on the case! Will close 😄

PR #238, opened 2018-02-07
Added type safety to bundle factory method

pablisco:
This PR makes the `Bundle` factory method (`bundleOf()`) more type safe. 

The following proposal should only produce compilation errors instead of runtime exceptions.

The negative tests (that resulted in exceptions) removed because it's not trivial to test a compilation error with the current set up (and probably shouldn't).


JakeWharton:
This is a lot of machinery. I suspect it's problematic for consumers since the stdlib `to` is imported by default. It also likely breaks usage of that function in files where you've explicitly imported.

I was hoping to write a lint check for this instead.

JakeWharton:
I don't think we want such a large API for this. It also means we need to contrive a replacement for `to` for every version of this for which there are least 3 so far (bundle, persistable bundle, content values). Discussion on better ways to handle this is in #263.

> Do we have support for library lint rules?

Yes. A sibling project houses the rules and then the `lintChecks` configuration on the main project automatically bundles them.

pablisco:
👍

PR #239, opened 2018-02-07
Add convenient `getBitmap` method for Resources and Context classes

rom4ek:
A couple of convenient methods to get rid of `BitmapFactory` annoyance. To get bitmap from resources in the same way as we get string from context:
```Kotlin
context.getString(R.string.something)
context.getBitmap(R.drawable.picture)
```


src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -111,3 +115,27 @@ inline fun createBitmap(
 ): Bitmap {
     return Bitmap.createBitmap(width, height, config, hasAlpha, colorSpace)
 }
+
+/**
+ * Return a [Bitmap] decoded from [Resources]
+ *
+ * @param resId Resource id of a drawable from `res/drawable` folder
+ * @param options null-ok; Options that control downsampling and whether the
```
romainguy: You don't need to specify null-ok, this information is embedded in the type

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -111,3 +115,27 @@ inline fun createBitmap(
...
+inline fun Resources.getBitmap(
+    @DrawableRes resId: Int,
+    options: BitmapFactory.Options? = null
+) = BitmapFactory.decodeResource(this, resId, options)
+
+/**
+ * Return a [Bitmap] decoded from [Resources] of this [Context]
+ *
+ * @param resId Resource id of a drawable from `res/drawable` folder
+ * @param options null-ok; Options that control downsampling and whether the
```
romainguy: Don't describe options in such detail, it's an incomplete description.

rom4ek:
Fixed. Do you think it makes sense to add the same kind of extension for `InputStream`, `FileDescriptor`, `ByteArray` and so on?

src/main/java/androidx/content/Context.kt
```diff
@@ -91,3 +95,27 @@ inline fun Context.withStyledAttributes(
...
+
+/**
+ * Return a [Bitmap] decoded from [Resources]
+ *
+ * @param resId Resource id of a drawable from `res/drawable` folder
+ * @param options null-ok; Options that control downsampling and whether the
+ * image should be completely decoded, or just is size returned.
+ */
+inline fun Resources.getBitmap(
+    @DrawableRes resId: Int,
```
NightlyNexus: it would be cool to have inline classes in the future: `DrawableInt.getBitmap(resources)` 😸

src/main/java/androidx/content/Context.kt
```diff
@@ -91,3 +95,27 @@ inline fun Context.withStyledAttributes(
...
+): Bitmap = BitmapFactory.decodeResource(this, resId, options)
+
+/**
+ * Return a [Bitmap] decoded from [Resources] of this [Context]
+ *
+ * @param resId Resource id of a drawable from `res/drawable` folder
+ * @param options null-ok; Options that control downsampling and whether the
+ * image should be completely decoded, or just is size returned.
+ */
+inline fun Context.getBitmap(
```
NightlyNexus: Are all Resources extension functions in this library going to have matching Context extension functions?

src/main/java/androidx/content/Context.kt
```diff
@@ -91,3 +95,25 @@ inline fun Context.withStyledAttributes(
...
     }
 }
+
+/**
+ * Return a [Bitmap] decoded from [Resources]
+ *
+ * @param resId Resource id of a drawable from `res/drawable` folder
+ * @param options Decoding options
+ */
+inline fun Resources.getBitmap(
```
NightlyNexus: would be cool when we get inline classes in the future: DrawableInt.getBitmap(resources) 😸

src/main/java/androidx/content/Context.kt
```diff
@@ -91,3 +95,25 @@ inline fun Context.withStyledAttributes(
...
+    options: BitmapFactory.Options? = null
+): Bitmap = BitmapFactory.decodeResource(this, resId, options)
+
+/**
+ * Return a [Bitmap] decoded from [Resources] of this [Context]
+ *
+ * @param resId Resource id of a drawable from `res/drawable` folder
+ * @param options Decoding options
+ */
+inline fun Context.getBitmap(
```
NightlyNexus: Are all Resources extension functions in this library going to have matching Context extension functions? (@jake?)

JakeWharton:
We can't merge this for a few reasons but the most important is that `decodeResource` handles only raster assets. This means that shape drawables, layer lists, insets drawables, vectors, and adaptive icons are unable to be decoded.

If you want a bitmap from a resource you're much better off loading it into a `Drawable` first and then calling our `toBitmap()` extension which will render it. This supports 100% of drawable types and also provides additional controls for the destination size.

Issue #193, opened 2018-02-06
TypedArray.use()


daberni:
```kotlin
inline fun TypedArray.use(block: (TypedArray) -> Unit) {
    try {
        block(this)
    } finally {
        recycle()
    }
}
```

for usage like this:

```kotlin
context.obtainStyledAttributes(attrs, R.styleable.TextView, 0, 0).use {
    val text = it.getString(R.styleable.TextView_text)
}
```

This would mimic the `Closeable` interface, see https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html

romainguy:
This was on my to-do list, thanks for the reminder :)

JakeWharton:
Do we need try/finally? A crash here _should_ take the entire process with it.

EddieRingle:
@JakeWharton I believe it should. If the intention is to mimic the Kotlin standard library's `use`, it should have the same semantics:

> Executes the given block function on this resource and _then closes it down correctly whether an exception is thrown or not_.

JakeWharton:
These are pooled objects. Not native file handles. I see no value in
introducing what is otherwise a code smell to maintain implementation
semantics to Kotlin's stdlib when we're merely trying to match usage
conventions. Simply omitting that clause from our documentation frees us
from it.

daberni:
If you are catching the exception on a higher level, the TypedArray isn't recycled. But I am not into the inner workings of TypedArray. When `recycle()` isn't necessary at all, please deprecate it and remove it from the official docs. But as long this isn't the case this extension would be handy.

JakeWharton:
You seem to be conflating best-effort pooling with actual leaks of resources. These are not equivalent.

JakeWharton:
Also we are not debating the value of the extension but its implementation.

daberni:
So it would be enough when replacing the `try-finally` block.

```
inline fun TypedArray.use(block: (TypedArray) -> Unit) {
    block(this)
    recycle()
}
```

JakeWharton:
That is what I would be an advocate for, yes.

JakeWharton:
Was merged.

PR #240, opened 2018-02-07
Add 'use' extension function to TypedArray, similar to Closeable's

DSteve595:
Addresses #193.

### DIFF
--- a/src/androidTest/java/androidx/content/res/TypedArrayTest.kt
+++ b/src/androidTest/java/androidx/content/res/TypedArrayTest.kt
@@ -204,4 +204,23 @@ class TypedArrayTest {
             array.getTextOrThrow(R.styleable.TypedArrayTypes_text_array_absent)
         }.hasMessageThat().isEqualTo("Attribute not defined in set.")
     }
+
+    @Test fun use() {
+        val attrs = context.getAttributeSet(R.layout.typed_array)
+        val array = context.obtainStyledAttributes(attrs, R.styleable.TypedArrayTypes)
+
+        var booleanPresent = false
+        array.use {
+            booleanPresent = it.getBooleanOrThrow(R.styleable.TypedArrayTypes_boolean_present)
+        }
+        assertTrue(
+            "Boolean obtained from attrs wasn't correct. " +
+                "Either the 'use' block wasn't run, or the boolean was missing from the TypedArray.",
+            booleanPresent
+        )
+
+        assertThrows<RuntimeException> {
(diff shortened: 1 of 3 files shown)


src/androidTest/java/androidx/content/res/TypedArrayTest.kt
```diff
@@ -204,4 +204,23 @@ class TypedArrayTest {
...
+        val attrs = context.getAttributeSet(R.layout.typed_array)
+        val array = context.obtainStyledAttributes(attrs, R.styleable.TypedArrayTypes)
+
+        var booleanPresent = false
+        array.use {
+            booleanPresent = it.getBooleanOrThrow(R.styleable.TypedArrayTypes_boolean_present)
+        }
+        assertTrue(
+            "Boolean obtained from attrs wasn't correct. " +
+                "Either the 'use' block wasn't run, or the boolean was missing from the TypedArray.",
```
romainguy: Since you call getBooleanOrThrow, isn't the only way to reach here that the block wasn't run?

src/androidTest/java/androidx/content/res/TypedArrayTest.kt
```diff
@@ -204,4 +204,23 @@ class TypedArrayTest {
...
+            booleanPresent
+        )
+
+        assertThrows<RuntimeException> {
+            array.recycle()
+        }.hasMessageThat().isEqualTo("$array recycled twice!")
```
romainguy: I don't think it's worth relying on an exact error message (esp. since it's not part of the public API and therefore not guaranteed).

DSteve595: Gotcha. Is there any way to check the value of `mRecycled` (a private field)?

romainguy: Only by reflection. The best thing to do here is just assert that a RuntimeException is thrown

DSteve595: Will update.

src/androidTest/java/androidx/content/res/TypedArrayTest.kt
DSteve595: Yep, my mistake! Will update.

DSteve595:
Addressed @romainguy's feedback.

JakeWharton:
Thanks! Squashed and merged. I changed the test to exercise the return value support of the method.

PR #248, opened 2018-02-07
-Simplify android.view.animation.Animation listener

jutikorn:
-tint color on Drawable
-set underLine and strikeThrough to TextView
-setMargins directly on View
-get child or null in ViewGroup

### DIFF
--- a/src/main/java/androidx/animation/Animation.kt
+++ b/src/main/java/androidx/animation/Animation.kt
@@ -34,7 +34,6 @@ fun Animation.doOnEnd(action: (animation: Animation) -> Unit) = addListener(onEn
  */
 fun Animation.doOnStart(action: (animation: Animation) -> Unit) = addListener(onStart = action)
 
-
 /**
  * Add an action which will be invoked when the animation has repeated.
  * @return the [Animation.AnimationListener] added to the Animation
@@ -65,4 +64,3 @@ fun Animation.addListener(
     setAnimationListener(listener)
     return listener
 }
-
--- a/src/main/java/androidx/animation/Animation.kt
+++ b/src/main/java/androidx/animation/Animation.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


romainguy:
Thanks for your submission but could you separate unrelated features in different PRs please?

src/main/java/androidx/animation/Animation.kt
```diff
@@ -0,0 +1,66 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.animation
+
+import android.view.animation.Animation
```
romainguy: android.view.Animation is the legacy animation system and not something we wish to extend further through this library.

src/main/java/androidx/view/TextView.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import android.graphics.Paint
+import android.widget.TextView
+
+/**
+ * Make an underline paintFlags to TextView
+ * */
+fun TextView.underLine() {
+    paintFlags = paintFlags or Paint.UNDERLINE_TEXT_FLAG
+}
```
romainguy: This would make more sense on spans, but not for an entire TextView.

src/main/java/androidx/view/View.kt
```diff
@@ -172,3 +173,19 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
 }
+
+/**
+ * Sets the margins in the ViewGroup's MarginLayoutParams. This version of the method sets all axes
+ * to the provided size.
+ *
+ * @see ViewGroup.MarginLayoutParams.setMargins
+ * @throws IllegalStateException if view doesn't have layoutParams as  ViewGroup.MarginLayoutParams
+ */
+fun View.setMargins(@Px size: Int) {
```
romainguy: This feature doesn't belong on View. A view should never query its layout params.

src/main/java/androidx/view/ViewGroup.kt
```diff
@@ -31,6 +31,15 @@ import android.view.ViewGroup
 operator fun ViewGroup.get(index: Int) =
     getChildAt(index) ?: throw IndexOutOfBoundsException("Index: $index, Size: $childCount")
 
+/**
+ * Returns the view at `index` or null
+ *
+ */
+fun ViewGroup.getOrNull(position: Int): View? = when {
```
romainguy: What's the use case for this?

jutikorn:
I'll make PR with separate feature.

PR #255, opened 2018-02-07
Add boolean getter extensions to Cursor

DavidHamm:
there is a way to put boolean into the sqlite database like

`db.insert("table", null, contentValuesOf("booleanColumn" to true))`

or

`db.update("table", contentValuesOf("booleanColumn" to true), null, null)`

but there is no native function to get a boolean from a cursor.

therefore it would be nice to have some boolean getter extensions like

`cursor.getBoolean("booleanColumn")`

or

`cursor.getBooleanOrNull("booleanColumn")`

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -65,6 +65,9 @@ package androidx.database {
     method public static final byte[] getBlob(android.database.Cursor, String columnName);
     method public static final byte[]? getBlobOrNull(android.database.Cursor, int index);
     method public static final error.NonExistentClass getBlobOrNull(android.database.Cursor, String columnName);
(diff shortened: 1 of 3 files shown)


DavidHamm:
I signed it!

davethomas11:
See #201 and #266 for opinions on adding extensions to Cursor or other database related code.

JakeWharton:
While this seems obvious, some DBs will store boolean values as T or F characters or literal values of "true"/"false" or "TRUE"/"FALSE". How you choose to store it is an implementation detail of your application and as such you can define these extensions directly in your project.

JakeWharton:
No. Sound like it should be facilitated by the framework.

PR #256, opened 2018-02-07
add 'isOdd', 'isEven' extension function to Int

tebitoq:
add extension to Int to know if a value is odd or even

```
val mValue = 3
mValue.isEven() -> false
mValue.isOdd() -> true
```

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -520,6 +520,12 @@ package androidx.util {
     method @RequiresApi(26) public static final android.util.Half toHalf(String);
   }
 
+  public final class IntKt {
+    ctor public IntKt();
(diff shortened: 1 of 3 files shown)


JakeWharton:
These extensions are not specific to Android. You should propose them to the Kotlin standard library instead or just retain them in your own codebase.

tebitoq:
Sorry my fault,  I understand your point but could you please explain why these extensions are valid ? just to be clear in future contributions.

https://github.com/android/android-ktx/blob/master/src/main/java/androidx/time/Year.kt

Regards!

JakeWharton:
Those types are at least defined as part of the Android SDK and we're
working on upstreaming them.

PR #257, opened 2018-02-07
Add SpannableString and SpannedString

hendraanggrian:
Conveniently get all spans with `SpannedString.kt`, and some operator methods in `SpannableString.kt`.

### DIFF
--- a/api/0.1.txt
+++ b/api/0.1.txt
@@ -343,6 +343,14 @@ package androidx.os {
 
 package androidx.text {
 
+  public final class SpannableStringKt {
+    ctor public SpannableStringKt();
+    method public static final void clearSpans(android.text.Spannable);
+    method public static final void minusAssign(android.text.Spannable, Object span);
+    method public static final void plusAssign(android.text.Spannable, Object span);
+    method public static final android.text.Spannable toSpannable(String);
+  }
+
   public final class SpannableStringBuilderKt {
     ctor public SpannableStringBuilderKt();
     method public static final android.text.SpannableStringBuilder backgroundColor(android.text.SpannableStringBuilder, @ColorInt int color, kotlin.jvm.functions.Function1<? super android.text.SpannableStringBuilder,kotlin.Unit> builderAction);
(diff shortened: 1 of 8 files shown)


api/0.1.txt
```diff
@@ -343,6 +343,14 @@ package androidx.os {
 
 package androidx.text {
 
+  public final class SpannableStringKt {
```
romainguy: Don't modify this file

hendraanggrian: I just reversed it. Are we supposed to change only the current.txt?

src/main/java/androidx/text/SpannableString.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import android.text.Spannable
+import android.text.SpannableString
+import android.text.SpannableStringBuilder
+import android.text.Spanned.SPAN_INCLUSIVE_EXCLUSIVE
+
+/**
+ * Returns a new [SpannableStringBuilder] from source,
+ * or the source itself if it is already an instance of [SpannableStringBuilder].
+ */
```
JakeWharton: Documentation doesn't match the return type

src/main/java/androidx/text/SpannedString.kt
```diff
@@ -0,0 +1,34 @@
...
+
+package androidx.text
+
+import android.text.Spanned
+import android.text.SpannedString
+
+/**
+ * Returns a new [SpannedString] from source,
+ * or the source itself if it is already an instance of [SpannedString].
+ */
```
JakeWharton: Documentation doesn't match the return type

src/main/java/androidx/text/SpannedString.kt
```diff
@@ -0,0 +1,34 @@
...
+inline fun CharSequence.toSpanned(): Spanned = SpannedString.valueOf(this)
+
+/** Returns `true` if [span] is found in this text. */
+inline operator fun Spanned.contains(span: Any): Boolean = getSpans<Any>().contains(span)
```
JakeWharton: I'm a bit worried about the overhead of this method. Is this operation really something that we think people do?

hendraanggrian: `Spanned.contains()` is probably not that common, I removed it in the latest commit. `CharSequence.toSpanned()` and `CharSequence.toSpannable()`, however, I imagine would be useful.

JakeWharton:
Merged manually. Thanks!

hendraanggrian:
Anytime!

PR #260, opened 2018-02-07
Add read/write properties for each view visibility.

### LINKED ISSUE #135: Visibility extensions for View classes
I'm considering adding a few methods to View class in order to set visibility, like: 
```
setVisible()
setInvisible()
setGone()
```
(API is a subject to change if you will)
What do you think guys? Is it something we would like to see in android-ktx?

JakeWharton:
Closes #135. Closes #146.

### DIFF
--- a/src/main/java/androidx/view/View.kt
+++ b/src/main/java/androidx/view/View.kt
@@ -172,3 +172,66 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
     }
     return Bitmap.createBitmap(width, height, config).applyCanvas(::draw)
 }
+
+/**
+ * Returns true when this view's visibility is [View.VISIBLE], false otherwise.
+ *
+ * ```
+ * if (view.isVisible) {
+ *     // Behavior...
+ * }
+ * ```
+ *
+ * Setting this property to true sets the visibility to [View.VISIBLE], false to [View.GONE].
+ *
+ * ```
+ * view.isVisible = true
+ * ```
+ */
+inline var View.isVisible: Boolean
+    get() = visibility == View.VISIBLE
+    set(value) {
+        visibility = if (value) View.VISIBLE else View.GONE
+    }
+
+/**
+ * Returns true when this view's visibility is [View.INVISIBLE], false otherwise.
+ *
+ * ```
+ * if (view.isInvisible) {
+ *     // Behavior...
+ * }
+ * ```
+ *
+ * Setting this property to true sets the visibility to [View.INVISIBLE], false to [View.VISIBLE].
+ *
+ * ```
+ * view.isInvisible = true
+ * ```
+ */
+inline var View.isInvisible: Boolean
+    get() = visibility == View.INVISIBLE
+    set(value) {
+        visibility = if (value) View.INVISIBLE else View.VISIBLE
+    }
+
+/**
+ * Returns true when this view's visibility is [View.GONE], false otherwise.
+ *
+ * ```
+ * if (view.isGone) {
+ *     // Behavior...
+ * }
+ * ```
+ *
+ * Setting this property to true sets the visibility to [View.GONE], false to [View.VISIBLE].
+ *
+ * ```
+ * view.isGone = true
+ * ```
+ */
+inline var View.isGone: Boolean
+    get() = visibility == View.GONE
+    set(value) {
+        visibility = if (value) View.GONE else View.VISIBLE
+    }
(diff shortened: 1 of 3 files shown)


src/main/java/androidx/view/View.kt
```diff
@@ -172,3 +172,66 @@ fun View.toBitmap(config: Bitmap.Config = Bitmap.Config.ARGB_8888): Bitmap {
...
+ * Setting this property to true sets the visibility to [View.VISIBLE], false to [View.GONE].
+ *
+ * ```
+ * view.isVisible = true
+ * ```
+ */
+inline var View.isVisible: Boolean
+    get() = visibility == View.VISIBLE
+    set(value) {
+        visibility = if (value) View.VISIBLE else View.GONE
```
LouisCAD: To me, the contrary of visible is invisible. Invisible means the thing is there, but not visible, which Android does right. Gone, means it's not there. Gone contrary would be more "present" rather than visible, so I think the else case should be `View.INVISIBLE`.

This is the way I did it in my WIP project, on purpose: https://github.com/LouisCAD/Splitties/blob/d9e733cce9a637c6f0f7e825ab9ae5c2be758934/views/src/main/java/splitties/views/Visibility.kt#L30

JakeWharton: I do not agree. I'm optimizing for the far more common case of visible/gone.

LouisCAD: IMHO, a [`showIf(yourBooleanExpression)`](https://github.com/LouisCAD/Splitties/blob/d9e733cce9a637c6f0f7e825ab9ae5c2be758934/views/src/main/java/splitties/views/Visibility.kt#L54) is as good for this more common case, and shouldn't confuse anyone, Android development experienced or not.

JakeWharton: Disparity with an entirely separate function using terminology inconsistent with the properties and constant names doesn't feel like a good alternative.

LouisCAD: Well, `showIf(bool)` extension is designed for the far more common case of visible/gone you mentioned, with a name that means shown with true, and not shown if false. You can also use `isGone` setter if it doesn't make you add an extra `!`.

For the less common use cases, I think using `isVisible` and `isInvisible` are great alternatives. Getting `isVisible` and variants is a code smell IMHO though, because I don't think having logic depending on View visibility is good design.

Having copy pasted the `visible`, `gone` and `showIf(bool)` extensions in several projects, I mostly used `showIf(bool)` and in some rare cases, I used `visible` setters for invisibility, but I never used them as getters, and when I think about the false case of `visible`, I think the View is not visible, so `View.INVISIBLE`.

An alternative would be to make a `shown` or `isShown` extension property to not have confusing false case of `isVisible` setter.

JakeWharton: I don't know what "not shown" means. Is that invisible or gone? So not only do you retain the ambiguity of behavior, you promote a completely different calling convention that can't handle the ternary nature of visibility.

This conversation is on further convincing me of the existing property design which provides uniformity to access and assignment and has the same documentation for disambiguating the inverse value that would be on any solution.

LouisCAD: While, in natural language, I qualify "not visible" and "invisible" to be the same, "shown" behavior may intriguate at first, so I'd check the documentation, that would explicitly state that "shown" is visible, and "not shown" is gone.

I'm not convinced that "not visible" should be "gone" when "invisible" is also a state, but I guess we have two different opinions on this specific case. Maybe other folks can give their take?

davethomas11: My most common use of the ```View.setVisibility``` function is toggling between the states Visible and Gone. 

The proposed property for ```visible``` mirrors what would be a useful utility method that I would include in an Android code base for convinience.

oldergod: Would it make sense to have something like this?
```kotlin
inline var View.isVisible = visibility == View.VISIBLE
inline fun View.setVisible = visibility = View.VISIBLE
```
`GONE` is definitely the more common case but reading code such as `view.isVisible = false`, or `isGone = false`, `isInvisible = false` would make me somewhat inconfortable.

JakeWharton: Val's only? Sure. I don't think `setVisible()` has any value over just calling `visibility = VISIBLE` though. And even the val barely passes my bar.

davethomas11: I don't know if I see the value in having the val only. It is not useful enough to constitute having it in the android-ktx code base. It is not very often that you have logic that is dependent on the visibility state of a View. I'd discourage code that is reliant on getting this value often enough that it needs a convenience method to do it. The common case is to be setting this value.

The var is more useful because usually changing this state is tied to some sort of boolean value in your logic. I'd argue that if it is going to be val don't consider this change at all.

PR #262, opened 2018-02-07
Add SpannableStringBuilder#scale

mightyfrog:
Add RelativeSizeSpan wrapper.


JakeWharton:
Is this common? We tried to include only the common ones as `inSpans` is always available for use with _any_ span.

mightyfrog:
I use it where bold doesn't have any effect, like some Japanese fonts, to have bold-ish effect by scaling up a little bit. Or simply per designer request. The most common use case among us is scaling up a number, like "Episode 99" on a label without using two TextViews.

mightyfrog:
RelativeSizeSpan is very useful together with ColorSpan to keep layout

mightyfrog:
Closed by mistake :(

JakeWharton:
merged manually. thanks!

PR #229, opened 2018-02-06
Add extensions for EditText.addTextChangedListener

sourabhv:

PR #269, opened 2018-02-07
TextView extensions

dovahkiin98:


dovahkiin98:
Signed in

JakeWharton:
Shouldn't these all be on `TextView`? `EditText` is an empty class.

dovahkiin98:
@JakeWharton I thought it would be more useful to have these methods on EditText since the getText() returns editable, and TextWatcher is used more on EditText

davethomas11:
Duplicates #151 #229

src/main/java/androidx/view/EditText.kt
```diff
@@ -0,0 +1,64 @@
...
+ * Set and Get the string value of EditText content
+ */
+var EditText.content
+    get() = text.toString()
+    set(value) {
+        setText(value)
+    }
+
+/**
+ * @returns Weather the EditText text is empty or not
```
pavs23: Do you mean "Weather"?

davethomas11: Rain or shine, the EditText is empty or not.

pavs23: Haha 😂

src/main/java/androidx/view/TextView.kt
```diff
@@ -0,0 +1,67 @@
...
+fun TextView.clear() {
+    text = ""
+}
+
+/**
+ * @param before action performed before text changes
+ * @param after action performed after text changes
+ * @param onTextChanged action preformed on text change
+ */
+fun TextView.onTextChanged(
```
JakeWharton: This should be `inline` and the lambdas should be `crossinline`.

src/main/java/androidx/view/TextView.kt
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * @returns Whether the EditText text is blank or not
+ */
+val TextView.isBlank get() = text.isBlank()
+
+/**
+ * Clears the contents of the EditText
+ */
+fun TextView.clear() {
```
JakeWharton: I'm not sure these 4 extensions have any real value. They're conveniences for things which are already pretty convenient.

79joey:
Can you please have some off your mobile application for my project at college

JakeWharton:
Text watcher-based listeners have been merged in http://r.android.com/720364. As I said above, I don't think the other extensions are worth including since they're so trivial and don't leverage any advantage of the language.

PR #274, opened 2018-02-07
String to ColorInt extension

### LINKED ISSUE #244: Extension on String to convert it to color
An extension like `String.toUri` to convert String to int color

```
inline fun String.toColor(): Int = Color.parseColor(this)
```

rom4ek:
Closes #244

### DIFF
--- a/src/main/java/androidx/graphics/Color.kt
+++ b/src/main/java/androidx/graphics/Color.kt
@@ -378,5 +378,18 @@ inline val @receiver:ColorLong Long.colorSpace: ColorSpace get() = Color.colorSp
 
 /**
  * Return a corresponding [Int] color of this [String].
+ *
+ * Supported formats are:
+ * ```
+ * #RRGGBB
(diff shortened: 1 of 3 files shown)


src/main/java/androidx/graphics/Color.kt
```diff
@@ -375,3 +375,8 @@ inline val @receiver:ColorLong Long.isWideGamut get() = Color.isWideGamut(this)
  */
 @get:RequiresApi(26)
 inline val @receiver:ColorLong Long.colorSpace: ColorSpace get() = Color.colorSpace(this)
+
+/**
+ * Return a corresponding [Int] color of this [String].
```
romainguy: This requires more documentation about the valid string formats.

rom4ek: Updated

JakeWharton:
Added `@ColorInt`, squashed, and merged.

PR #278, opened 2018-02-08
Add strike through to SpannableStringBuilder

jutikorn:
add a strike Through to SpannableStringBuilder with `StrikethroughSpan()`


JakeWharton:
Merged manually. Thanks!

Issue #235, opened 2018-02-06
checkApi and updateApi ignore methods with reified generics


dewarder:
Following method is ignored by `checkApi` as well as by `updateApi`:

```kotlin
inline fun <reified T> doSomething() {
}
```

As follows `Context.systemService` is absent in `current.txt`.

JakeWharton:
If memory serves, this is a UAST bug.

tnorbye:
Finally got around to investigating this, and I'm realizing that it's not actually a UAST bug. 

It looks like what the Kotlin compiler does with reified methods like that is to turn them private!
If I access your method as a Kotlin extension method from Kotlin, it works and compiles:
context.systemService<String>() // Compiles

However, if I try to access it from Java as a util method, it fails:
        ContextKt.systemService(context); // Doesn't compile
with this message from javac:
error: <T>systemService(Context) has private access in ContextKt
where T is a type-variable:
T extends Object declared in method <T>systemService(Context)

So, the AST marks this method as private. And when metalava is visiting the API of the class, it sees that the method isn't public and therefore drops it.

Until now I've made the metalava signature files reflect the Java-observable API (e.g. a property exploded into its getters and setters etc). But this is a pretty clear example of a case where it needs to diverge from that and actually list the Kotlin-only-accessible part of the API. I'm thinking I should just switch the signatures over (for Kotlin-compilation-units) to more directly reflect Kotlin. The main downside to this is when Java code is replaced by equivalent Kotlin code; it will be a big visible change in the signatures of the class, which it really shouldn't be. I'm going to think about it, but please chime in if you have strong opinions one way or the other!

JakeWharton:
It sounds like we might need to track both projections of the API. Not
here, where we only care about the Kotlin view, but for regular support
libraries.

For example, the difference between a getter function and a property
defined in Kotlin is very different call-site semantics when called from
Kotlin. From the Java side, though, both are the same getter method.

PR #281, opened 2018-02-08
Add ViewGroup extension to allow concisely modifying of the layoutParams

steurt:
Added ViewGroup extension to allow concisely modifying of the layoutParams.

This allows following usage:
```
viewGroup.layoutParamsAs<ViewGroup.MarginLayoutParams> {
    updateMarginsRelative(start = marginStart)
}
```

Instead of:
```
val marginLayoutParams = viewGroup.layoutParams as ViewGroup.MarginLayoutParams
marginLayoutParams.updateMarginsRelative(start = 16)
viewGroup.layoutParams = marginLayoutParams
```

Was unable to update the `api` due to reified generics issue https://github.com/android/android-ktx/issues/235.

### DIFF
--- a/src/main/java/androidx/view/ViewGroup.kt
+++ b/src/main/java/androidx/view/ViewGroup.kt
@@ -78,6 +78,21 @@ val ViewGroup.children: Sequence<View>
     }
 
 /**
+ * Executes [block] with a typed version of the ViewGroup's layoutParams and reassigns the
+ * layoutParams with the updated version.
+ *
+ * @see ViewGroup.getLayoutParams
+ * @see ViewGroup.setLayoutParams
+ **/
+inline fun <reified T> ViewGroup.layoutParamsAs(
+    block: T.() -> Unit
+) where T : ViewGroup.LayoutParams {
+    val typedLayoutParams = layoutParams as T
+    block(typedLayoutParams)
+    layoutParams = typedLayoutParams
+}
+
+/**
  * Sets the margins in the ViewGroup's MarginLayoutParams. This version of the method sets all axes
  * to the provided size.
  *
--- a/src/main/java/androidx/view/ViewGroup.kt
+++ b/src/main/java/androidx/view/ViewGroup.kt
@@ -78,13 +78,23 @@ val ViewGroup.children: Sequence<View>
     }
 
 /**
+ * Executes [block] with the layoutParams and reassigns the layoutParams with the updated version.
+ *
+ * @see ViewGroup.getLayoutParams
+ * @see ViewGroup.setLayoutParams
+ **/
+inline fun ViewGroup.updateLayoutParams(block: ViewGroup.LayoutParams.() -> Unit) {
+    updateLayoutParamsAs(block)
+}
+
+/**
  * Executes [block] with a typed version of the ViewGroup's layoutParams and reassigns the
  * layoutParams with the updated version.
  *
  * @see ViewGroup.getLayoutParams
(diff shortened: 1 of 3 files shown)


steurt:
I signed it!

steurt:
I signed it!

steurt:
I signed it!

steurt:
I signed it again!

src/main/java/androidx/view/ViewGroup.kt
```diff
@@ -77,6 +77,21 @@ val ViewGroup.children: Sequence<View>
...
     }
 
+/**
+ * Executes [block] with a typed version of the ViewGroup's layoutParams and reassigns the
+ * layoutParams with the updated version.
+ *
+ * @see ViewGroup.getLayoutParams
+ * @see ViewGroup.setLayoutParams
+ **/
+inline fun <reified T> ViewGroup.layoutParamsAs(
```
romainguy: The name does not match the behavior. I would expect this method to simply be a get+cast, not to trigger a requestLayout via setting the LPs.

steurt: Do you think `withLayoutParams` would be better? Suggestions?

MGaetan89: `updateLayoutParams`?

steurt: @romainguy would `updateLayoutParams` or `updateLayoutParamsAs` work for you?

steurt:
Renamed from `layoutParamsAs` to `updateLayoutParamsAs`. The usage then becomes:
```
viewGroup.updateLayoutParamsAs<ViewGroup.MarginLayoutParams> {
    updateMarginsRelative(start = marginStart)
}
```

Added non-generic variant `updateLayoutParams` for simply updating the layout params as `ViewGroup.LayoutParams`. This allows following usage:
```
viewGroup.updateLayoutParams { width = 500 }
```

steurt:
Renamed from `layoutParamsAs` to `updateLayoutParamsAs`. The usage then becomes:
```
viewGroup.updateLayoutParamsAs<ViewGroup.MarginLayoutParams> {
    updateMarginsRelative(start = marginStart)
}
```

Added non-generic variant `updateLayoutParams` for simply updating the layout params as `ViewGroup.LayoutParams`. This allows following usage:
```
viewGroup.updateLayoutParams { width = 500 }
```

steurt:
@romainguy any thoughts?

steurt:
I can name the generic version `updateLayoutParams` as well. However, in doing so the non-generic and generic variant will have the same JVM signature, i.e. a platform declaration clash.

This can be solved by specifying a different `@JvmName` annotation for each.
Perhaps:
```
@JvmName("updateViewGroupLayoutParams")
inline fun ViewGroup.updateLayoutParams(block: ViewGroup.LayoutParams.() -> Unit)

@JvmName("updateLayoutParams")
inline fun <reified T> ViewGroup.updateLayoutParams(block: T.() -> Unit)
```

Will this do?

romainguy:
Should we even have the non generic version at this point? @JakeWharton?

JakeWharton:
Playing with it locally.

JakeWharton:
How about this? https://github.com/android/android-ktx/compare/steurt-master

steurt:
@JakeWharton seems like a good setup to me!

romainguy:
LGTM

PR #285, opened 2018-02-08
Get UUID for database Cursor

jspetrak:


JakeWharton:
This is an app-specific use case and not a general one. Moreover, you should probably be using something like Room or SQL Delight for storing non-native scalar types in the database.

PR #289, opened 2018-02-08
Feature/log

VitaminPSG:
Logs usage:

- Equivalent of  `Log.i(javaClass.simpleName, message)`

```
override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.test_activity)
        i("onCreate")
    }
```
```
   i {
            when (Build.VERSION.SDK_INT) {
                Build.VERSION_CODES.HONEYCOMB -> "HONEYCOMB"
                Build.VERSION_CODES.FROYO -> "FROYO"
                else -> "UNKNOWN"
            }
       }
```
```
i("Custom tag", "Log my version ${Build.VERSION.SDK_INT}")
```

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -523,35 +523,22 @@ package androidx.util {
   public final class LogKt {
     ctor public LogKt();
     method public static final <T> void d(T, kotlin.jvm.functions.Function0<String> msg);
-    method public static final <T> void d(T, String msg, java.lang.Object... args);
     method public static final <T> void d(T, String msg);
-    method public static final void d(String tag, String msg, java.lang.Object... args);
     method public static final int d(String tag, String msg);
     method public static final <T> void e(T, kotlin.jvm.functions.Function0<String> msg);
-    method public static final <T> void e(T, String msg, java.lang.Object... args);
     method public static final <T> void e(T, String msg);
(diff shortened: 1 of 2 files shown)


sourabhv:
I'd prefer it as 

```
info("I have a snake in my boot")
error("To infinity and ... StackOverflowException")
wtf("WTF! I still have a snake in my boot")
```

And the implementation could be

```
inline fun Any.info(message: String) = Log.i(javaClass.simpleName, message)
```

But I don't find this very useful

romainguy:
The `isLoggable` aren't super useful since at that point you've already allocated all your strings.

lukaspili:
Many kotlin developer create their own log() function, which calls `Timber.d` rather than `Log.d`.
Not sure it's such a great idea to pollute the IDE import window with an identical `log()` function that you don't want to use.
What about adding a prefix like `androidDebug()`, `androidError()`, etc?

JakeWharton:
I don't think this provides a lot of added value. It's mostly code golfing character count. The lazy string one is interesting to potentially add as static extensions onto `Log` that check `isLoggable`, but we can't do that until https://youtrack.jetbrains.com/issue/KT-11968 is fixed so going to close for now.

PR #290, opened 2018-02-08
Toast extensions on Context

simonschiller:
See #143


simonschiller:
I signed it!

build.gradle
```diff
@@ -63,6 +63,7 @@ dependencies {
 
     androidTestImplementation 'junit:junit:4.12'
     androidTestImplementation "com.android.support:support-annotations:${versions.supportLibrary}"
+    androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.1'
```
shaishavgandhi05: Extract version to project build.gradle?

simonschiller: Currently, only versions that are used multiple times are extracted (`kotlin`, `supportLibrary` and `atsl`), I wanted to keep consistent with that.

romainguy: Before we merge this, @JakeWharton do you have any problem with adding espresso?

JakeWharton: Yep. It's a bit overkill though as we could inspect the properties of the returned `Toast` object.

src/main/java/androidx/widget/Toast.kt
```diff
@@ -0,0 +1,40 @@
...
+import android.content.Context
+import android.support.annotation.StringRes
+import android.widget.Toast
+
+/**
+ * Creates and shows a [Toast] with the given [text]
+ *
+ * @param duration Toast duration, defaults to [Toast.LENGTH_SHORT]
+ */
+fun Context.toast(text: CharSequence, duration: Int = Toast.LENGTH_SHORT): Toast {
```
JakeWharton: inline

src/main/java/androidx/widget/Toast.kt
```diff
@@ -0,0 +1,40 @@
...
+    return Toast.makeText(this.applicationContext, text, duration).apply { show() }
+}
+
+/**
+ * Creates and shows a [Toast] with text from a resource
+ *
+ * @param resId Resource id of the string resource to use
+ * @param duration Toast duration, defaults to [Toast.LENGTH_SHORT]
+ */
+fun Context.toast(@StringRes resId: Int, duration: Int = Toast.LENGTH_SHORT): Toast {
```
JakeWharton: inline

src/main/java/androidx/widget/Toast.kt
```diff
@@ -0,0 +1,40 @@
...
+    return Toast.makeText(this.applicationContext, resId, duration).apply { show() }
```
JakeWharton: Why the application context? Just use `this`.

src/main/java/androidx/widget/Toast.kt
```diff
@@ -0,0 +1,40 @@
...
+import android.support.annotation.StringRes
+import android.widget.Toast
+
+/**
+ * Creates and shows a [Toast] with the given [text]
+ *
+ * @param duration Toast duration, defaults to [Toast.LENGTH_SHORT]
+ */
+fun Context.toast(text: CharSequence, duration: Int = Toast.LENGTH_SHORT): Toast {
+    return Toast.makeText(this.applicationContext, text, duration).apply { show() }
```
JakeWharton: Why the application context? Just use `this`.

romainguy: Doh, good catch. Yeah don't use the application context.

src/main/java/androidx/widget/Toast.kt
romainguy: Toast doesn't have a `getText()` only a `getView()`, which could make the inspection brittle.

JakeWharton: Espresso is doing a DFS for TextView's whose text matches the supplied string which would be pretty easy to replicate.

romainguy: 👍

simonschiller:
Updated the PR @JakeWharton @romainguy

JakeWharton:
Looks great. Thanks!

PR #291, opened 2018-02-08
Api version check methods

rom4ek:
A couple of convenient methods to execute some actions for specific api versions

```Kotlin
fromApi(21, {
    // Do something...
})
```

Not sure about inclusive/exclusive cases though. Should they be supplied as arguments?

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -317,6 +317,12 @@ package androidx.net {
 
 package androidx.os {
 
+  public final class BuildKt {
+    ctor public BuildKt();
+    method public static final void fromApi(int fromVersion, kotlin.jvm.functions.Function0<kotlin.Unit> block);
+    method public static final void toApi(int toVersion, kotlin.jvm.functions.Function0<kotlin.Unit> block);
+  }
+
   public final class BundleKt {
     ctor public BundleKt();
     method public static final error.NonExistentClass bundleOf(kotlin.Pair<String,?>... pairs);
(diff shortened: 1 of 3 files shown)


JakeWharton:
I don't find these names particularly intuitive. And I'm not convinced there's value here over an `if`. Why use these?

simonschiller:
Maybe `untilApi` would be better than `toApi`.

rom4ek:
The naming could be better, something like `sinceSdkVersion` and `beforeSdkVersion`. Why use them - cause it gets annoying and takes me quite some time occasionally to recall the whole `Build.VERSION.SDK_INT` line.

JakeWharton:
But isn't knowing that these methods exist and recalling their name equivalent?

rom4ek:
Well, yes, but this is more about auto-complete thingy. When I start typing `fromApi` or whatever it's called - I'll get the wanted function, which I could already use, whilst typing `Build` or `VERSION` still requires some recalling/search of following `SDK_INT`.

JakeWharton:
I don't think I buy that argument because you have to know what the
function is called to get auto complete just like you need to know Build.
Wouldn't you also expect people to use the Build constants for the API
values anyway?

And finally, this doesn't allow any structure like else, else if, or when.
So if you need complex control flow you have to totally change your calling
convention from the lambda back to normal if or when.

rom4ek:
Yes, the latter argument totally makes sense. Closing then.

PR #297, opened 2018-02-09
Add extension for RemoteViews

HeyAlex:
Add extension for RemoteViews. Just add most popular methods with @Remotable annotation. 
PR with tests.

Usage: 
`remoteView.setEnabled(R.id.test_imageView, false)`
`remoteView.setBackgroundResource(R.id.test_textView, R.drawable.box)`

### DIFF
--- a/src/androidTest/java/androidx/testing.kt
+++ b/src/androidTest/java/androidx/testing.kt
@@ -18,12 +18,17 @@ package androidx
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.graphics.Bitmap
+import android.graphics.drawable.Drawable
 import android.support.annotation.LayoutRes
 import android.util.AttributeSet
 import android.util.Xml
+import androidx.graphics.drawable.toBitmap
 import com.google.common.truth.ThrowableSubject
 import com.google.common.truth.Truth.assertThat
 import org.xmlpull.v1.XmlPullParser
+import java.io.ByteArrayOutputStream
+import java.util.Arrays
 
 inline fun <reified T : Throwable> assertThrows(body: () -> Unit): ThrowableSubject {
     try {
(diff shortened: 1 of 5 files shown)


JakeWharton:
I don't know anything about this. So i'll have to learn before merging...

HeyAlex:
@JakeWharton RemoteViews is intended for custom notifications and widgets. So view's methods that have annotation (@RemotableViewMethod) can be invoked in RemoteViews, but there is no any methods for that. So the only way to invoke these methods is to use setBooleans/setInt/setDouble and name of the method, and kotlin extensions is perfectly solve this problem.

JakeWharton:
Yeah I've used remote views with widgets a bunch. Where's the documentation for these strings though? How do I know this is the complete subset?

HeyAlex:
@JakeWharton As i understand where is no documentation for this thing. So if you want to set imageView resource, RemoteViews contains:
```java
    /**
     * Equivalent to calling ImageView.setImageResource
     *
     * @param viewId The id of the view whose drawable should change
     * @param srcId The new resource id for the drawable
     */
    public void setImageViewResource(int viewId, int srcId) {
        setInt(viewId, "setImageResource", srcId);
    }
```

But if you want to set alpha for that image, you need to make:
```java
setInt(viewId, "setImageAlpha", alpha)
```

 The way is to grep code by annotation (@remotableviewmethod) (something like that [here](http://grepcode.com/search/usages?id=repository.grepcode.com%24java%24ext@com.google.android%24android@5.1.1_r1@android%24view@RemotableViewMethod&type=type&k=u)). I just add the most popular methods, so it's not a complete subset (don't know, should i add them all, cuz it isn't that popular case).

NightlyNexus:
Is there an existing ticket to get these methods added to the Android SDK itself?
Does this belong in a different support library like v4 compat?

HeyAlex:
Actually i didnt find any related issues on their tracker. Seems like it might be in support v4 compat as you said. @JakeWharton  Should i close this PR, cuz i still don't know what the best approach, or it's still suitable for ktx?.

PR #303, opened 2018-02-10
Fix Lint warnings

simonschiller:
Fixes almost all Lint warnings (#292)

Only one left is for the outdated dependency `com.android.tools.build:gradle:3.1.0-beta1`, I am not really sure what we should do here. Should we ignore the inspection (probably a bad idea) or should we update to `com.android.tools.build:gradle:3.2.0-alpha01`?


jaredsburrows:
Thanks!

build.gradle
```diff
@@ -72,12 +72,19 @@ dependencies {
 
     // Metalava isn't released yet. Check in its jar and explicitly track its transitive deps.
     metalava files('metalava.jar')
+    // noinspection GradleDependency
```
jaredsburrows: Ah, I am not sure if you want to do this. Just turn the inspection off in your own IDE.

romainguy: Jake has a PR to disable lint warnings. I'd like to get the fixes for TypedArrayTest.kt checked in though. Could you remove the changes to build.gradle please?

JakeWharton: The test changes aren't really needed since it just leaks recycled objects out of their pool, not native resource handles that actually accumulate. The GC grabs the instances when they go out of scope and the pool will replenish itself as it gets drained.

PR #304, opened 2018-02-10
Add extension for ArrayList

pokk:
Added a extension of ArrayList for `removeRange`.

The useage is as the following:

```kotlin
listArray.removeRange(2, 4)
```


MGaetan89:
This is not Android-specific, so it should rather be included in Kotlin stdlib.

PR #305, opened 2018-02-10
Convenience function for retrieving styled & formatted string resources

simonschiller:
Usefull when you want to retrieve a string resource which contains styling information and also acts as a format string. See [String Resources Documentation](https://developer.android.com/guide/topics/resources/string-resource.html#StylingWithHTML).


simonschiller:
Is this PR worth keeping or should I close it?
@JakeWharton @romainguy

PR #306, opened 2018-02-10
Add delegate extension for three reference type

pokk:
Added three delegate classes for `SoftReference`, `WeakReference`, and `PhantomReference`.

The usage as the following:

```kotlin
val variable by SoftRef(TestData())
val variable by WeakRef(TestData())
val variable by PhantomRef(TestData())
```

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -96,6 +96,30 @@ package androidx.database.sqlite {
 
 }
 
+package androidx.delegate {
+
+  public final class PhantomRef<T> implements kotlin.properties.ReadWriteProperty<java.lang.Object,T> {
+    ctor public PhantomRef(T? p, java.lang.ref.ReferenceQueue<T> queue);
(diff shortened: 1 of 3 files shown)


MGaetan89:
This is not Android-specific, so it should rather be included in Kotlin stdlib.

pokk:
@MGaetan89 

Thank you for mentioning me that kindly.
But so sorry for that I don't get it, there are few extensions in the `util` folder, as like, `ArraySet.kt`, ... I think they're the same as my pr if I don't misunderstand.
Could you please tell me what's the difference?

MGaetan89:
`ArraySet` is in the `android.util` package, and is, as such, an Android provided class.
The class you want to "improve" are in the `java.lang` package. They are not only available in Android, but in Java in general.

pokk:
Thank you for your replying and spending your time for my pr. I'll close this.

PR #309, opened 2018-02-10
Add view locationOnScreen/locationInWindow extensions

rom4ek:
Add `locationOnScreen`/`locationInWindow` extensions for more convenient usage

```Kotlin
val (x, y) = view.locationOnScreen
val point = view.locationOnScreen
```


rom4ek:
Thanks for you response. But why the whole `time` package which requires api 26 is in this library then? Given the percantage of devices with Oreo installed, I'm pretty sure `locationOnScreen` will be more usable.

PR #311, opened 2018-02-11
Use inline modifier to Animator listeners

kubode:
Non-inline functions has overhead. So I fix it by adding crossinline parameters defaulted with empty lambda.


kubode:
🤔
You're exactly right.
I'll close this PR. Thank you for reviewing!

PR #313, opened 2018-02-11
Add `MarginLayoutParams` extensions for destructuring

rom4ek:
Having the idea that `MarginLayoutParams` are commonly used for dealing with margins, I thought it might be useful to have destructuring extensions for this, so it could be used like:
```Kotlin
val (left, top, right, bottom) = marginLayoutParams
```

I'm not sure how to deal with `marginEnd`/`marginStart` though. Should we have an sdk version check in place, to decide whether we should return `marginStart` or `leftMargin` ?

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -648,6 +648,10 @@ package androidx.view {
 
   public final class ViewGroupKt {
     ctor public ViewGroupKt();
+    method public static final operator int component1(android.view.ViewGroup.MarginLayoutParams);
+    method public static final operator int component2(android.view.ViewGroup.MarginLayoutParams);
+    method public static final operator int component3(android.view.ViewGroup.MarginLayoutParams);
+    method public static final operator int component4(android.view.ViewGroup.MarginLayoutParams);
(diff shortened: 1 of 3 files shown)


romainguy:
My issue with this is it could be confusing with subclasses (many layouts subclass from `MarginLayoutParams`). Either the destructuring is limited – always – to margins, or you should have a least 4 destructuring components in the subclasses (otherwise you'd get a mix of margins and something else).

rom4ek:
I would restrict destructuring to margins only. But is it doable in Kotlin now?

romainguy:
Not that I know of. Also you would thus prevent custom layouts to be able to do their own destructuring. I like this extension but I'm not convinced it's that useful to be able to destructure the margins.

rom4ek:
Ok, I got your point. What do you think on destructuring into width/height for `ViewGroup.LayoutParams` ?

romainguy:
I don't think we should add destructuring to layout params. I can't think of a meaningful way of doing it since they are subclasses so frequently.

PR #318, opened 2018-02-12
Kotlin commons and tests

gmjgh:
- Kotlin commons extensions for Long, String, Boolean and Any
- Tests for general/Commons
- API changes

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -96,6 +96,19 @@ package androidx.database.sqlite {
 
 }
 
+package androidx.general {
+
+  public final class CommonsKt {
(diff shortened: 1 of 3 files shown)


JakeWharton:
This library is focused on the Android APIs only. These should be requested upstream on the Kotlin standard library.

gmjgh:
@JakeWharton does this apply to support libraries like design or cardview?

JakeWharton:
No. Just the platform.

gmjgh:
I signed it! (by the way) :)

PR #332, opened 2018-02-16
Added some extension functions fro work with toast and adding fragments.

aztek-god:


JakeWharton:
Platform fragments are deprecated in P so we won't be adding any extensions for them. This library only covers the platform so we won't be adding extensions for support fragments. There's already discussions about toast extensions in #143 and an initial PR at #290. Also this PR lacks tests and you didn't update the API files.

PR #333, opened 2018-02-16
Extension for AppCompatActivity for doing fragment transactions

themanol:


JakeWharton:
This project is extensions only for the platform. In the future we might have extensions for other support libraries.

PR #338, opened 2018-02-19
Add an extension for adding an animator listener by DSL

pokk:
It can add an animator listener by DSL as following below

```kotlin
animator.addAnimatorListener {
	onEnd { ... }
	onStart { ... }
	onCancel { ... }
	onRepeat { ... }
}
```

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -2,6 +2,7 @@ package androidx.animation {
 
   public final class AnimatorKt {
     ctor public AnimatorKt();
+    method public static final error.NonExistentClass addAnimatorListener(android.animation.Animator, kotlin.jvm.functions.Function1<? super androidx.animation.WrapperAnimatorListener,kotlin.Unit> listener);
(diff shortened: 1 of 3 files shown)


JakeWharton:
We already have individual animation listeners and we don't want both. I don't see a lot of motivation to switch to this style since the likelihood that you're applying more than one callback is low and the overhead of doing so in the current style isn't terrible. Am I missing anything here?

pokk:
Thank you for your feedback. I just though using DSL might be better.
And I agree with you that only for the DSL and plus a class is not a good idea. Is there a way to create the DSL without extra a class?

hendraanggrian:
The thing about this is, if we start adding such feature, other listeners need to be converted to DSL as well. And there's a lot of them: `TextWatcher`, `OnGestureListener`, `OnScrollListener`, `OnItemSelectedListener`, `OnQueryTextListener`, `OnDrawerScrollListener`, etc.

pokk:
@hendraanggrian 
Thank you for your suggestion. I think they will be a lot, too.

Issue #301, opened 2018-02-10
doBackground / Easier async task


geniushkg:
Hey team,
Thanks for this effort, I appreciate your hard work in making developer's life easier.

New Feature : Do in Background using coroutines 
doInBg{ .. }

Similar implementation in Anko [Anko coroutines](https://github.com/Kotlin/anko/wiki/Anko-Coroutines)

I got the inspiration from #196 which was closed, so this issue will be focused on
 AsyncTask/Background operation using coroutines.

Please add your valuable suggestion.

illuzor:
Coroutines is separate library, which must be connected to this library.

Maybe something like [this](https://github.com/illuzor/KAndroid/blob/master/kandroid/src/main/kotlin/com/pawegio/kandroid/KThread.kt)?

geniushkg:
@illuzor, Kandroid under hood uses threads, anyway thanks for the reference.

SjoerdvGestel:
@geniushkg  or anko-coroutines from JetBrains?
https://github.com/Kotlin/anko

Muhannad508:
Hey @geniushkg , please tell me what do u think about this implementation . I have tested it, and it works fine

```
val execute = AsyncTaskKTX<Int, Int, Int>().runInBackground {
            doInBg {
                println("doInBg function in test , Thread : ${Thread.currentThread().name}")
                Assert.assertEquals(4, it.size)
                for (i in 1..5) {
                    Thread.sleep(1000)
                    publishProgress(i)
                }
                it.size
            }

            onUpdate {
                println("onUpdate (${it[0]}) function in test , Thread : ${Thread.currentThread().name}")

            }
        }.execute(3, 4, 5, 6)
```
where AsyncTaskKTX is extend of AsyncTask .

`class AsyncTaskKTX<Params, Progress, Result> : AsyncTask<Params, Progress, Result>()`

Muhannad508:
Hey @geniushkg , if you are still working on implementing this feature, then take a look at this article below. hope it helps you avoid all the limitations/problems that already exist in AsyncTask. 

http://blog.danlew.net/2014/06/21/the-hidden-pitfalls-of-asynctask/

PR #341, opened 2018-02-19
Async task implementation #301

Muhannad508:
an easier way to use async task without the need of extending AsyncTask . see #301 
Usage: 
```
val execute = AsyncTaskKTX<Int, Int, Int>().runInBackground {
        doInBg {
            //executed in background
        }
        onUpdate {

        }
        onPostExecute {

        }
    }.execute(3, 4, 5, 6)

```


Muhannad508:
I signed it!

JakeWharton:
You've mixed the `gh-pages` and `master` branch together. Code should be done on the `master` branch only.

PR #342, opened 2018-02-19
easier way to use AsyncTask.

Muhannad508:
easier way to use async task without the need of extending AsyncTask . see #301 
Usage:

```
val execute = AsyncTaskKTX<Int, Int, Int>().runInBackground {
        doInBg {
            //executed in background
        }
        onUpdate {

        }
        onPostExecute {

        }
    }.execute(3, 4, 5, 6)

```

### DIFF
--- a/src/androidTest/java/androidx/os/AsyncTaskTest.kt
+++ b/src/androidTest/java/androidx/os/AsyncTaskTest.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


src/androidTest/java/androidx/os/AsyncTaskTest.kt
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Created by muhannad on 2/16/18.
+ */
+class AsyncTaskTest {
+
+    @Test
+    fun testtask() {
+        println("main , Thread : ${Thread.currentThread().name}")
+        val execute = AsyncTaskKTX<Int, Int, Int>().runInBackground {
```
romainguy: I don't like this API: you create an object, then call two methods on it. I'm not sure what this buys you compared to just creating an object that overrides the appropriate methods? At the very least creating the asynctask could be done in runInBackground.

Speaking of which, names: AsynctaskKTX isn't a good name and runInBackground seems to indicate that it will run the task, but it doesn't.

JakeWharton:
I don't think we've established the need for this API fully. Let's keep conversation in #301.

PR #343, opened 2018-02-19
Simple factory method for creating LruCache without subclassing it

svasilinets:


src/androidTest/java/androidx/util/LruCacheTest.kt
```diff
@@ -0,0 +1,35 @@
+package androidx.util
```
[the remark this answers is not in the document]
svasilinets: Done.

src/main/java/androidx/util/LruCache.kt
```diff
@@ -0,0 +1,41 @@
...
+ * Returns the computed value or null if no value can be computed. The default implementation
+ * returns null.
+ * @param onEntryRemoved a function called for entries that have been evicted or removed.
+ *
+ * @see LruCache.sizeOf
+ * @see LruCache.create
+ * @see LruCache.entryRemoved
+ */
+fun <K, V> lruCache(maxSize: Int,
+                    sizeFunc: (K, V) -> Int = { _, _ -> 1 },
```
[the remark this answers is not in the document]
svasilinets: Done.

src/main/java/androidx/util/LruCache.kt
```diff
@@ -0,0 +1,57 @@
...
+ * @param onCreate a onCreate called after a cache miss to compute a value for the corresponding key.
+ * Returns the computed value or null if no value can be computed. The default implementation
+ * returns null.
+ * @param onEntryRemoved a function called for entries that have been evicted or removed.
+ *
+ * @see LruCache.sizeOf
+ * @see LruCache.create
+ * @see LruCache.entryRemoved
+ */
+fun <K, V> lruCache(maxSize: Int,
```
JakeWharton: We should make this `inline` and have the lambdas be `crossinline` and use an anonymous type rather than a named one. This will only allocate a single class rather than one per lambda.

JakeWharton: Both `K` and `V` should be restricted to `: Any` as they do not allow nulls

src/main/java/androidx/util/LruCache.kt
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (C) 2017 The Android Open Source Project
```
JakeWharton: 2018?

src/main/java/androidx/util/LruCache.kt
```diff
@@ -0,0 +1,57 @@
...
+ * @param onEntryRemoved a function called for entries that have been evicted or removed.
+ *
+ * @see LruCache.sizeOf
+ * @see LruCache.create
+ * @see LruCache.entryRemoved
+ */
+fun <K, V> lruCache(maxSize: Int,
+                    onSizeOf: (K, V) -> Int = { _, _ -> 1 },
+                    onCreate: (K) -> V? = { _ -> null },
+                    onEntryRemoved: ((Boolean, K, V, V?) -> Unit) = { _, _, _, _ -> }):
```
JakeWharton: Can we add names to the lambda arguments which match the argument names of the originating class? We did a pass before open sourcing to make sure they were all there. I haven't been doing a good job of making sure new code has them...

JakeWharton:
Squashed and merged (and formatted). Thanks!

PR #367, opened 2018-02-27
Add AccountManager.removeAccount() that works on all platforms

jkreiser:
In API 22 (LOLLIPOP_MR1),
AccountManager.removeAccount(Account, Activity, AccountManagerCallback<Bundle>, Handler) was added, and
AccountManager.removeAccount(Account, AccountManagerCallback<Boolean>, Handler) was deprecated.

To resolve this when running on a pre-22 device, call the old removeAccount(), get the Boolean result,
put it in a Bundle using key AccountManager.KEY_BOOLEAN_RESULT,
then call AccountManagerCallback<Bundle> callback.run() with a new AccountManagerFuture<Bundle>
that returns this Bundle.

Credit also to my co-workers @joshfriend and @scottschmitz at @MichiganLabs for their contributions.

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -1,3 +1,12 @@
+package androidx.accounts {
+
+  public final class AccountManagerExt {
+    ctor public AccountManagerExt();
+    method public static void removeAccount(android.accounts.AccountManager, android.accounts.Account account, android.app.Activity? activity = "null", android.os.Handler? handler = "null", kotlin.jvm.functions.Function1<? super android.accounts.AccountManagerFuture<android.os.Bundle>,kotlin.Unit>? callback = "null");
+  }
+
+}
+
 package androidx.animation {
 
   public final class AnimatorKt {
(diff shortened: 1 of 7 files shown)


JakeWharton:
This seems like something you should [propose for inclusion into the support library][1]. Nothing about this leverages something unique to Kotlin except for the fact that it's inside an extension method. But what would be more appropriate is for the support library to contain this functionality and then a KTX library for the support library placing the helper onto the `AccountManager` via an extension.

 [1]: https://issuetracker.google.com/issues/new?component=192731&template=842428

jkreiser:
[I've opened up an issue for its inclusion into the support library.](https://issuetracker.google.com/issues/111961751)  My apologies for the delay.

godeadasdf:
2018年7月30日 22:46，Joseph Kreiser <notifications@github.com>写道：I've opened up an issue for its inclusion into the support library.. My apologies for the delay.

PR #370, opened 2018-02-27
Add DisplayMetrics and its test

hendraanggrian:
Easy dp, sp conversion to pixel and vice versa.


JakeWharton:
This doesn't address any of the concerns in #132.

PR #379, opened 2018-03-01
Adding shrek asset

AlecStrong:
Should help


jrodbx:
LGTM!

JakeWharton:
This work is under copyright that you do not have permission to give to us for use. If you hand draw a version of this on a napkin and take a high resolution scan, however, that will be sufficiently derivative and could be merged.

PR #384, opened 2018-03-02
Kotlin 1.2.30 updates

jaredsburrows:
Previous: https://github.com/android/android-ktx/pull/380

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:3.1.0-beta3'
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:${versions.kotlin}"
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$versions.kotlin"
         classpath "org.jetbrains.dokka:dokka-android-gradle-plugin:0.9.16"
     }
 }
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -79,7 +79,7 @@ dependencies {
     metalava 'com.android.tools.lint:lint-checks:26.1.0-beta1'
     metalava 'com.android.tools.lint:lint-gradle:26.1.0-beta1'
     metalava 'com.android.tools.lint:lint:26.1.0-beta1'
-    metalava 'org.jetbrains.kotlin:kotlin-stdlib:1.2.21'
+    metalava "org.jetbrains.kotlin:kotlin-stdlib:$versions.kotlin"
```
jaredsburrows: This was missing the `$versions.kotlin` constant.

build.gradle
```diff
@@ -14,7 +14,7 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:3.1.0-beta3'
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:${versions.kotlin}"
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$versions.kotlin"
```
jaredsburrows: `{}` are not needed.

build.gradle
JakeWharton: This version is specifically pinned to the version metalava uses.

JakeWharton:
I prefer `{` and `}` for expression templates (like Kotlin requires) because its far easier to read.

PR #385, opened 2018-03-02
Extension for Html.fromHtml - closes #317

### LINKED ISSUE #317: CharSequence / String fromHtml()
What do you think about such extension?

```kotlin
fun CharSequence.fromHtml(flags: Int? = null): Spanned {
    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
        Html.fromHtml(this.toString(), flags ?: Html.FROM_HTML_MODE_LEGACY)
    } else {
        Html.fromHtml(this.toString())
    }
}
```

jaredsburrows:
Closes #317 @JakeWharton @romainguy


src/main/java/androidx/text/Html.kt
```diff
@@ -0,0 +1,51 @@
...
+import android.os.Build
+import android.text.Html
+import android.text.Html.ImageGetter
+import android.text.Html.TagHandler
+import android.text.Spanned
+
+/**
+ * Returns a new [Spanned] from [CharSequence] HTML.
+ */
+fun CharSequence.parseAsHtml(flags: Int? = 0): Spanned {
```
jaredsburrows: `flags: Int? = 0` because https://developer.android.com/reference/android/text/Html.html#FROM_HTML_MODE_LEGACY.

romainguy: Could you please explain this behavior in the documentation of the method? It might worth it for users to understand what the API will do on different versions of the platform.

src/main/java/androidx/text/Html.kt
```diff
@@ -0,0 +1,51 @@
...
+        @Suppress("DEPRECATION")
+        Html.fromHtml(toString())
+    }
+}
+
+/**
+ * Returns a new [Spanned] from [CharSequence] HTML.
+ */
+fun CharSequence.parseAsHtml(
+    flags: Int? = 0,
```
romainguy: Same thing here

jaredsburrows: `flags: Int? = 0` explain this?

romainguy: Yep, explain what the default is and why, and what happens when invoking parseAsHtml on N+ and before N.

jaredsburrows: Thanks.

jaredsburrows: Fixed!

jaredsburrows:
@romainguy Documentation updated!

src/main/java/androidx/text/Html.kt
```diff
@@ -0,0 +1,66 @@
...
+import android.os.Build
+import android.text.Html
+import android.text.Html.ImageGetter
+import android.text.Html.TagHandler
+import android.text.Spanned
+
+/**
+ * Returns a new [Spanned] from [CharSequence] HTML.
+ *
+ * In Android 4.4, additional options were introduced to allow for more
```
romainguy: This does not match the implementation. Android N was 7.0, not 4.4.

jaredsburrows: Fixed. Bad copy-pasta.

jaredsburrows:
Anything else you want me to fix in this PR @romainguy? :)

src/androidTest/java/androidx/text/HtmlTest.kt
```diff
@@ -0,0 +1,61 @@
...
+    @Test fun parseAsHtmlFlagsImageGetterTagHandler() {
+        val expected = "Hi \u00a9 > <"
+        val actual = "<b>Hi</b> © > <"
+            .parseAsHtml(imageGetter = imageGetter, tagHandler = tagHandler)
+            .toString()
+
+        assertEquals(expected, actual)
+    }
+
+    private val imageGetter = ImageGetter { null }
```
romainguy: Nitpick: properties should be at the top of the class

jaredsburrows: This has been moved to the top. Anything else before merging? :)

jaredsburrows:
Anything else you want me to fix in this PR @romainguy? :)

jaredsburrows:
Can we merge? @romainguy? :)

jaredsburrows:
@romainguy Thank you!

PR #387, opened 2018-03-02
Add View.announceForAccessibility with String res as parameter

ataulm:
There's an API on View which allows announcements passing `CharSequence`, but not a `@StringResource` - this just delegates.

Not sure how to add a test for this one, but happy to try if you have ideas.

### DIFF
--- a/src/main/java/androidx/view/View.kt
+++ b/src/main/java/androidx/view/View.kt
@@ -25,6 +25,7 @@ import android.support.annotation.StringRes
 import android.support.v4.view.ViewCompat
 import android.view.View
 import android.view.ViewTreeObserver
+import android.view.accessibility.AccessibilityEvent
 import androidx.graphics.applyCanvas
 
 /**
@@ -86,7 +87,7 @@ inline fun View.doOnPreDraw(crossinline action: (view: View) -> Unit) {
 }
 
 /**
- * Sends {@link AccessibilityEvent} of type {@link AccessibilityEvent#TYPE_ANNOUNCEMENT}.
+ * Sends [AccessibilityEvent] of type [AccessibilityEvent.TYPE_ANNOUNCEMENT].
  *
  * @see View.announceForAccessibility
  */
--- a/src/main/java/androidx/view/View.kt
+++ b/src/main/java/androidx/view/View.kt
@@ -21,6 +21,7 @@ package androidx.view
 import android.graphics.Bitmap
 import android.support.annotation.Px
 import android.support.annotation.RequiresApi
+import android.support.annotation.StringRes
 import android.support.v4.view.ViewCompat
 import android.view.View
 import android.view.ViewTreeObserver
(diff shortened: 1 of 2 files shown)


src/main/java/androidx/view/View.kt
```diff
@@ -84,6 +85,17 @@ inline fun View.doOnPreDraw(crossinline action: (view: View) -> Unit) {
     })
 }
 
+/**
+ * Sends {@link AccessibilityEvent} of type {@link AccessibilityEvent#TYPE_ANNOUNCEMENT}.
+ *
+ * @see View.announceForAccessibility
+ */
+@RequiresApi(16)
```
ataulm: I didn't use `Build.VERSION_CODES.JELLY_BEAN` to match the style of the other functions in this file.

src/main/java/androidx/view/View.kt
```diff
@@ -84,6 +85,17 @@ inline fun View.doOnPreDraw(crossinline action: (view: View) -> Unit) {
     })
 }
 
+/**
+ * Sends {@link AccessibilityEvent} of type {@link AccessibilityEvent#TYPE_ANNOUNCEMENT}.
```
jaredsburrows: Make sure you use Dokka instead of JavaDocs.

You will need these imports:
```
import android.view.accessibility.AccessibilityEvent
import android.view.accessibility.AccessibilityEvent.TYPE_ANNOUNCEMENT
```

And change `Sends {@link AccessibilityEvent} of type {@link AccessibilityEvent#TYPE_ANNOUNCEMENT}.` to `Sends [AccessibilityEvent] of type [AccessibilityEvent.TYPE_ANNOUNCEMENT].`.

ataulm: Thanks!

JakeWharton:
> Not sure how to add a test for this one, but happy to try if you have ideas.

My quick guess would be to attach an accessibility delegate to a view, send the announcement, and then verify that an event was sent with https://developer.android.com/reference/android/view/accessibility/AccessibilityEvent.html#TYPE_ANNOUNCEMENT.

Want to give that a try?

ataulm:
I will give that a try tomorrow 👌

ataulm:
Unless an accessibility service is running on the device, it'll be a no-op 😞 

```
    public void announceForAccessibility(CharSequence text) {
        if (AccessibilityManager.getInstance(mContext).isEnabled() && mParent != null) {
            AccessibilityEvent event = AccessibilityEvent.obtain(
                    AccessibilityEvent.TYPE_ANNOUNCEMENT);
            onInitializeAccessibilityEvent(event);
            event.getText().add(text);
            event.setContentDescription(null);
            mParent.requestSendAccessibilityEvent(this, event);
        }
    }
```
I can enable/disable accessibility services programmatically but it requires allowing `WRITE_SECURE_SETTINGS` over adb before the tests run, which I guess we don't want to do 😅

JakeWharton:
Oh, well. It's a trivial implementation! Thanks.

PR #392, opened 2018-03-03
Fix various Kotlin issues

jaredsburrows:
Fixed various Kotlin issues separated by commits.
@JakeWharton @romainguy


src/main/java/androidx/view/Menu.kt
```diff
@@ -30,12 +30,7 @@ inline operator fun Menu.get(index: Int): MenuItem = getItem(index)
 
 /** Returns `true` if [item] is found in this menu. */
 operator fun Menu.contains(item: MenuItem): Boolean {
-    for (index in 0 until size) {
-        if (getItem(index) == item) {
-            return true
-        }
-    }
-    return false
+    return (0 until size).any { getItem(it) == item }
```
NightlyNexus: this isn't the same in bytecode, btw. this will create an IntRange (at least as of 1.2.21). the for loop is optimized by the compiler.

jaredsburrows: Revert this then? The Kotlin plugin made this suggestion.

romainguy: Yes, the original code used a loop on purpose

jaredsburrows: I will review and suppress for single statement.

jaredsburrows: This has now been fixed!

jaredsburrows:
Anything else @NightlyNexus and @romainguy?

PR #393, opened 2018-03-03
Add missing tests

jaredsburrows:


jaredsburrows:
I believe there are more we can do as well.

JakeWharton:
I added this method to P so the extension is about to become uncallable. Let's remove it.

jaredsburrows:
@JakeWharton What is P? You want me to remove `removeAt`?

Apsaliya:
Android P.

PR #401, opened 2018-03-05
Add SpannableString.overlay and .overlayAll

shekibobo:
The overlay methods will apply a provided span to matching
substrings of the receiver. In `.overlay`, it will apply to
only the first matching occurrence, while `.overlayAll` will
apply the span to all matching occurrences of the substring.

Because spans are not copyable, you provide a `spanBuilder` lambda
to generate a new span to apply to each occurrence.

This supports case-insensitive matching.


src/main/java/androidx/text/SpannableString.kt
```diff
@@ -70,3 +70,59 @@ inline operator fun Spannable.set(range: IntRange, span: Any) {
...
+ */
+inline fun Spannable.overlayAll(
+    substring: String,
+    ignoreCase: Boolean = false,
+    startIndex: Int = 0,
+    crossinline spanBuilder: () -> Any
+) {
+    generateSequence(startIndex) { overlay(substring, spanBuilder(), ignoreCase, it) }
+        .takeWhile { length >= it || it != -1 }
+        .toList()
```
JakeWharton: A sequence seems overkill. You suffer from boxing on all the integers, allocation of the intermediate operator wrappers, and allocation of the list which is thrown away.

Let's do a simple `while` loop instead?

src/main/java/androidx/text/SpannableString.kt
```diff
@@ -70,3 +70,59 @@ inline operator fun Spannable.set(range: IntRange, span: Any) {
...
+ * ```
+ *
+ * Note: Does nothing if the substring is not found.
+ *
+ * @param ignoreCase true to ignore character case when matching a string. By default false.
+ * @param startIndex starting index to search the string. Default: 0
+ *
+ * @see Spannable.setSpan
+ */
+inline fun Spannable.overlayAll(
```
JakeWharton: I don't think this should be `inline` as it's not a simple alias of one or two function calls.

src/main/java/androidx/text/SpannableString.kt
```diff
@@ -70,3 +70,59 @@ inline operator fun Spannable.set(range: IntRange, span: Any) {
...
+    substring: String,
+    ignoreCase: Boolean = false,
+    startIndex: Int = 0,
+    crossinline spanBuilder: () -> Any
```
JakeWharton: Can we call this something other than builder? spanFactory? spanCreator?

shekibobo: Going with `spanFactory`, and updated the docs to reflect the span comes from that instead of incorrectly indicating `span` as a parameter.

src/main/java/androidx/text/SpannableString.kt
shekibobo: Wasn't sure the rule about using `inline`. Seemed like everything was using it, so I just stuck with it where it didn't cause problems. Just a regular `fun` now.

src/main/java/androidx/text/SpannableString.kt
shekibobo: All good things to learn, thanks! Using a simple while loop now, and also fixed the halting problem with using `||` instead of `&&` for multiple negative conditions 🤦‍♂️ .

shekibobo:
Thanks! I just expanded the tests to get a little more coverage and a bit more specificity of each case. We now test the following for both methods:

- No matches
- Multiple matches
- Case sensitivity
- Start index
- Multiple different spans applied successfully

JakeWharton:
This is great. Thank you.

My biggest concern here is that the only thing Kotlin-specific is the fact that this is an extension function instead of a static method. I'm going to try to get this into the normal support library somewhere and then we can just hang it as an extension and alias.

shekibobo:
Okay, sounds awesome, thanks!

PR #403, opened 2018-03-06
Get File from Uri

### LINKED ISSUE #397: Get file from Uri
Can we get File from android Uri.
Something like var file = uri.toFile()

Sathawale27:
Closes #397


Sathawale27:
I signed it!

src/main/java/androidx/os/File.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import android.net.Uri
+import java.io.File
+
+/**
+ * Creates a File from the given Uri.
+ *
+ * @see File
+ */
+inline fun Uri.toFile(): File = File(path)
```
JakeWharton: This should be in `Uri.kt` in `android/net/`

src/androidTest/java/androidx/os/FileTest.kt
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package androidx.os
+
+import android.net.Uri
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import java.io.File
+
+class FileTest {
```
JakeWharton: There's already `UriTest` where this should live.

src/main/java/androidx/os/File.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package androidx.os
+
+import android.net.Uri
+import java.io.File
+
+/**
+ * Creates a File from the given Uri.
+ *
+ * @see File
```
JakeWharton: Delete

src/main/java/androidx/os/File.kt
```diff
@@ -0,0 +1,29 @@
...
+
+@file:Suppress("NOTHING_TO_INLINE") // Aliases to public API.
+
+package androidx.os
+
+import android.net.Uri
+import java.io.File
+
+/**
+ * Creates a File from the given Uri.
```
JakeWharton: Please use braces `[ .. ]` around `File` and `Uri` as they are types.

Sathawale27:
It returns the File so i thought it would be in File.kt. May i know why instrumented test is getting failed?
Thanks @JakeWharton.

JakeWharton:
They're broken for external contributors with no easy fix in sight.

PR #406, opened 2018-03-07
Base 64 Kotlin extensions

jaredsburrows:


src/main/java/androidx/util/Base64.kt
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+package androidx.util
+
+import android.util.Base64
+import java.nio.charset.Charset
+
+/**
+ * From [String] to Base-64 encoded [String].
```
JakeWharton: This is a sentence fragment. Prefix it with a verb like "Convert".

src/main/java/androidx/util/Base64.kt
```diff
@@ -0,0 +1,56 @@
...
+ */
+fun String.toBase64(
+    offset: Int = 0,
+    len: Int = this.length,
+    flags: Int = Base64.DEFAULT
+): String = Base64.encode(toByteArray(), offset, len, flags).toString(Charset.forName("US-ASCII"))
```
JakeWharton: The toByteArray charset parameter should be hoisted onto this function as we cannot assume UTF-8 is the desired encoding. And below.

JakeWharton: Use http://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-charsets/-u-s_-a-s-c-i-i.html instead of looking up charset by name. And below.

src/main/java/androidx/util/Base64.kt
JakeWharton: Also nit: lowercase "Base" on all these

src/main/java/androidx/util/Base64.kt
```diff
@@ -0,0 +1,56 @@
...
+fun ByteArray.toBase64(
+    offset: Int = 0,
+    len: Int = this.size,
+    flags: Int = Base64.DEFAULT
+): ByteArray = Base64.encode(this, offset, len, flags)
+
+/**
+ * From Base-64 encoded [String] to decoded [String].
+ */
+fun String.fromBase64(
```
JakeWharton: The Kotlin standard library typically puts "from" methods as extensions on the companion of the type on which they operate and take an instance as the first parameter.

src/main/java/androidx/util/Base64.kt
```diff
@@ -0,0 +1,56 @@
...
+    offset: Int = 0,
+    len: Int = this.length,
+    flags: Int = Base64.DEFAULT
+): String = Base64.decode(toByteArray(), offset, len, flags).toString(Charset.forName("US-ASCII"))
+
+/**
+ * From Base-64 encoded [ByteArray] to decoded [ByteArray].
+ */
+fun ByteArray.fromBase64(
```
JakeWharton: These can all be `inline` since they're aliases.

src/androidTest/java/androidx/util/Base64Test.kt
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.Test
+
+class Base64Test {
+    @Test fun toBase64String() {
+        val expected = "SGksIEphcmVk"
+        val actual = "Hi, Jared".toBase64(flags = Base64.NO_WRAP)
+        assertEquals(expected, actual)
+    }
+
+    @Test fun fromBase64String() {
```
JakeWharton: pair of byte array tests too?

src/main/java/androidx/util/Base64.kt
jaredsburrows: There is no `companion object` in `ByteArray`.

jaredsburrows:
@JakeWharton All fixed. What do you want to do about `ByteArray` not having a `companion object`?

JakeWharton:
We might have to leave it out until https://youtrack.jetbrains.com/issue/KT-11968 is fixed. I'll ask JetBrains about it at our weekly meeting next week.

src/main/java/androidx/util/Base64.kt
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Convert [String] to base-64 encoded [String].
+ */
+inline fun String.toBase64(
+    offset: Int = 0,
+    len: Int = this.length,
+    flags: Int = Base64.DEFAULT,
+    charset: Charset = Charsets.US_ASCII
+): String = Base64.encode(toByteArray(), offset, len, flags).toString(charset)
```
JakeWharton: The charset should be used for `toByteArray` and the default should be `UTF_8`. The ASCII charset is for `toString()`.

src/main/java/androidx/util/Base64.kt
```diff
@@ -0,0 +1,61 @@
...
+
+package androidx.util
+
+import android.util.Base64
+import java.nio.charset.Charset
+
+/**
+ * Convert [String] to base-64 encoded [String].
+ */
+inline fun String.toBase64(
```
JakeWharton: Let's call this `.encodeBase64` as there's no `Base64` type so we're not really going "to" it.

src/main/java/androidx/util/Base64.kt
```diff
@@ -0,0 +1,61 @@
...
+inline fun ByteArray.toBase64(
+    offset: Int = 0,
+    len: Int = this.size,
+    flags: Int = Base64.DEFAULT
+): ByteArray = Base64.encode(this, offset, len, flags)
+
+/**
+ * Convert base-64 encoded [String] to decoded [String].
+ */
+inline fun String.Companion.fromBase64(
```
JakeWharton: This can go back to being an instance extension and be called `decodeBase64()`.

src/main/java/androidx/util/Base64.kt
```diff
@@ -0,0 +1,61 @@
...
+    input: String,
+    offset: Int = 0,
+    len: Int = input.length,
+    flags: Int = Base64.DEFAULT,
+    charset: Charset = Charsets.US_ASCII
+): String = Base64.decode(input.toByteArray(), offset, len, flags).toString(charset)
```
JakeWharton: Pass `US_ASCII` to `toByteArray`.

src/main/java/androidx/util/Base64.kt
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Convert base-64 encoded [String] to decoded [String].
+ */
+inline fun String.Companion.fromBase64(
+    input: String,
+    offset: Int = 0,
+    len: Int = input.length,
+    flags: Int = Base64.DEFAULT,
+    charset: Charset = Charsets.US_ASCII
```
JakeWharton: Should be `UTF_8`

src/main/java/androidx/util/Base64.kt
```diff
@@ -0,0 +1,61 @@
...
+
+import android.util.Base64
+import java.nio.charset.Charset
+
+/**
+ * Convert [String] to base-64 encoded [String].
+ */
+inline fun String.toBase64(
+    offset: Int = 0,
+    len: Int = this.length,
```
JakeWharton: Length won't work here as you are using it as the length of the bytes whereas the parameter represents the number of characters. You need to do a `substring` before calling `toByteArray` and then pass the length of the byte array to `Base64`.

I would also write a test for this with two multi-byte UTF-8 characters in the source string and then a length of 1 which should thus encode more than one byte.

jaredsburrows: I updated this PR with everything else.

How would this substring work?

jaredsburrows: Any update on how you want this to work? Or do you mind if I just make the PR for the String to Base64 and Base64 to String?

PR #408, opened 2018-03-07
Add CharSequence.presence()

shekibobo:
This is similar to the `String#presence` method in Ruby/Rails. Returns
the string/charsequence if it has non-whitespace characters in it, otherwise 
return null.

Useful for providing easy fallbacks when things might be null *or*
blank, and you want to have a non-blank string provided.

```kotlin
val name = getNameFromSomewhereItCouldBeAnything()
userName.text = name.presence() ?: "Unknown"
```

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -351,6 +351,7 @@ package androidx.text {
   public final class CharSequenceKt {
     ctor public CharSequenceKt();
     method public static boolean isDigitsOnly(CharSequence);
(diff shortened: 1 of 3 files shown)


JakeWharton:
This isn't specific to Android and doesn't use an Android API in its implementation. You can suggest general functionality to the Kotlin standard library instead.

JakeWharton:
Also worth noting that you can do this with things like `takeIf` with a few more characters.

PR #409, opened 2018-03-07
Added View.toggleVisibility() extension

Link184:
I think will be nice to have a mechanism to easily change view visibility =)


JakeWharton:
We already have this in the mutable properties `isVisible`, `isInvisible`, and `isGone` which each toggle between two of the ternary states. You can use the one which corresponds to your desired pair and read/write it with a boolean.

PR #413, opened 2018-03-08
Add test for View.announceForAccessibility

ataulm:
As per: https://github.com/android/android-ktx/pull/387

I'm not sure if you wanted the literal in the assertion - I chose to resolve it in case the value is changed in the resources; we don't care about this specific value, we care that the method is called with the `CharSequence` that corresponds to the given `StringRes`.

I'm also not 100% on how much value this test gives - it solely tests that we delegate correctly, but I guess that's all the original function does :thinking:


JakeWharton:
Thanks for following up.

PR #415, opened 2018-03-08
Remove parseAsHtml overloading, add toHtml.

JakeWharton:
I just landed an HtmlCompat into the support library which we can delegate to directly in a future version.

Closes #404

### DIFF
--- a/src/androidTest/java/androidx/text/HtmlTest.kt
+++ b/src/androidTest/java/androidx/text/HtmlTest.kt
@@ -16,10 +16,11 @@
 
 package androidx.text
 
-import android.text.Html
+import android.text.Html.FROM_HTML_MODE_COMPACT
 import android.text.Html.ImageGetter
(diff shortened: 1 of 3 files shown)


src/androidTest/java/androidx/text/HtmlTest.kt
```diff
@@ -16,45 +16,47 @@
...
+    }
 
-        assertEquals(expected, actual)
+    @Test fun convertToHtml() {
+        val html = buildSpannedString {
+            bold {
+                append("Hi")
+            }
+        }.toHtml()
+        assertTrue(html, html.contains("<b>Hi</b>"))
```
JakeWharton: There's other tags which are added to this which I'm not convinced are stable across all API levels so we just check for the part we know will be there.

jaredsburrows:
Thanks for fix!

PR #417, opened 2018-03-09
Add `startActivity` and `startService` extension

hendraanggrian:
It shorten

```kotlin
startActivity(Intent(this, NextActivity::class.java)
    .putExtra(EXTRA_STRING, "Hello world")
    .putExtra(EXTRA_INT, 12))
```

to `startActivity<NextActivity>(EXTRA_STRING to "Hello world", EXTRA_INT to 12)`.

Although I'm not that sure these are necessary, what do you think?

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -14,6 +14,18 @@ package androidx.animation {
 
 }
 
+package androidx.app {
+
+  public final class ActivityKt {
+    ctor public ActivityKt();
+  }
+
+  public final class FragmentKt {
+    ctor public FragmentKt();
+  }
+
+}
+
 package androidx.content {
 
   public final class ContentValuesKt {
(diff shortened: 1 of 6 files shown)


JakeWharton:
See this comment: https://github.com/android/android-ktx/pull/189#issuecomment-371317100. We don't need functions for this because they're already trivial. We just need a better way to create intents.

hendraanggrian:
What about treating it like a bundle: `intentOf<NextActivity>(EXTRA_STRING to "Hello world")`.

JakeWharton:
That's a start, sure, but we also have to handle the cases where you don't
have a class instance and where you want to set category, data, etc.
Intents are complex structures.

hendraanggrian:
In that case, custom initialization block should suffice:

```kotlin
intentOf<NextActivity>(key to value) {
    setData(someData)
    addFlags(SOME_FLAG)
    setCategory(SOME_CATEGORY)
}
```

I'm counting 2 functions (with and without reified generics).

PR #418, opened 2018-03-09
Bitmap extensions

feiyanke:
some useful functions of Bitmap:
``` 
Bitmap.toBytes
Bitmap.clip
Bitmap.skew
Bitmap.rotate
```

### DIFF
--- a/src/main/java/androidx/graphics/Bitmap.kt
+++ b/src/main/java/androidx/graphics/Bitmap.kt
@@ -19,10 +19,18 @@
 package androidx.graphics
 
 import android.graphics.Bitmap
+import android.graphics.Bitmap.CompressFormat
+import android.graphics.Bitmap.CompressFormat.JPEG
+import android.graphics.Bitmap.CompressFormat.PNG
+import android.graphics.Bitmap.createBitmap
+import android.graphics.BitmapFactory
 import android.graphics.Canvas
 import android.graphics.ColorSpace
+import android.graphics.Matrix
 import android.support.annotation.ColorInt
 import android.support.annotation.RequiresApi
+import java.io.ByteArrayOutputStream
+import java.io.File
 
 /**
  * Creates a new [Canvas] to draw on this bitmap and executes the specified
@@ -111,3 +119,49 @@ inline fun createBitmap(
 ): Bitmap {
     return Bitmap.createBitmap(width, height, config, hasAlpha, colorSpace)
 }
+
+
+/**
+ * Bitmap to bytes.
+ *
+ * @param format The format of bitmap.
+ * @param quality Hint to the compressor, 0-100. 0 meaning compress for small size, 100 meaning compress for max quality.
+ * @return ByteArray
+ */
+fun Bitmap.toBytes(format:CompressFormat = JPEG, quality: Int=100):ByteArray
+        = ByteArrayOutputStream().also { compress(format, quality, it) }.toByteArray()
+
+/**
+ * Return the clipped bitmap.
+ *
+ * @param x      The x coordinate of the first pixel.
+ * @param y      The y coordinate of the first pixel.
+ * @param width  The width.
+ * @param height The height.
+ * @return the clipped bitmap
+ */
+fun Bitmap.clip(x:Int, y:Int, width: Int, height: Int):Bitmap
+        = Bitmap.createBitmap(this, x, y, width, height)
+
+/**
+ * Return the skewed bitmap.
+ *
+ * @param kx  The skew factor of x.
+ * @param ky  The skew factor of y.
+ * @param px  The x coordinate of the pivot point.
+ * @param py  The y coordinate of the pivot point.
+ * @return the skewed bitmap
+ */
+fun Bitmap.skew(kx:Float, ky:Float, px:Float = 0f, py:Float = 0f):Bitmap
+        = createBitmap(this, 0, 0 , width, height, Matrix().apply { skew(kx, ky, px, py) }, true)
+
+/**
+ * Return the rotated bitmap.
+ *
+ * @param degrees The number of degrees.
+ * @param px      The x coordinate of the pivot point.
+ * @param py      The y coordinate of the pivot point.
+ * @return the rotated bitmap
+ */
+fun Bitmap.rotate(degrees:Int, px: Float, py: Float):Bitmap
+        = createBitmap(this, 0, 0, width, height, Matrix().apply { rotate(degrees, px, py) }, true)


feiyanke:
I signed it!

hendraanggrian:
I'm liking `toBytes`, but it really should be called `toByteArray`. Also, I'm not sure if JPG shoud be default format over PNG.

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -111,3 +119,49 @@ inline fun createBitmap(
...
+
+
+/**
+ * Bitmap to bytes.
+ *
+ * @param format The format of bitmap.
+ * @param quality Hint to the compressor, 0-100. 0 meaning compress for small size, 100 meaning compress for max quality.
+ * @return ByteArray
+ */
+fun Bitmap.toBytes(format:CompressFormat = JPEG, quality: Int=100):ByteArray
```
romainguy: Compressing a bitmap can be done in many ways and compressing into a byte array is just a highly specific use case. I'd rather remove this method.

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -111,3 +119,49 @@ inline fun createBitmap(
...
+/**
+ * Return the clipped bitmap.
+ *
+ * @param x      The x coordinate of the first pixel.
+ * @param y      The y coordinate of the first pixel.
+ * @param width  The width.
+ * @param height The height.
+ * @return the clipped bitmap
+ */
+fun Bitmap.clip(x:Int, y:Int, width: Int, height: Int):Bitmap
```
romainguy: I'm not sure this method is necessary. It doesn't save much over the `Bitmap.createBitmap()` call for an operation that might not be very common. At least it should be `inline`, better documented and the return type doesn't need to be specified.

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -111,3 +119,49 @@ inline fun createBitmap(
...
+/**
+ * Return the skewed bitmap.
+ *
+ * @param kx  The skew factor of x.
+ * @param ky  The skew factor of y.
+ * @param px  The x coordinate of the pivot point.
+ * @param py  The y coordinate of the pivot point.
+ * @return the skewed bitmap
+ */
+fun Bitmap.skew(kx:Float, ky:Float, px:Float = 0f, py:Float = 0f):Bitmap
```
romainguy: This is not a common usage, please remove.

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -111,3 +119,49 @@ inline fun createBitmap(
...
+
+/**
+ * Return the rotated bitmap.
+ *
+ * @param degrees The number of degrees.
+ * @param px      The x coordinate of the pivot point.
+ * @param py      The y coordinate of the pivot point.
+ * @return the rotated bitmap
+ */
+fun Bitmap.rotate(degrees:Int, px: Float, py: Float):Bitmap
```
romainguy: I'm not entirely convinced by how common this use case is. (Also same comments that I've mentioned on `clip()` apply here).

feiyanke:
@hendraanggrian It more make sense to use `toByteArray`, I have fixed it and added `IntRange` for `quality`. PNG is a common use format, but I perfer JPG as it could make smaller size for most bitmaps.

src/main/java/androidx/graphics/Bitmap.kt
feiyanke: `clip` is of more clear meaning than `createBitmap` for bitmap operation. I added `inline` for all methods and updated docs.

src/main/java/androidx/graphics/Bitmap.kt
feiyanke: I perfer add this method in order to let operations of bitmap more complete

api/current.txt
```diff
@@ -23,6 +23,7 @@ package androidx.content {
 
   public final class ContextKt {
     ctor public ContextKt();
+    method public static java.util.List<java.lang.String> getPermissions(android.content.Context);
```
romainguy: Why this change?

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -111,3 +116,56 @@ inline fun createBitmap(
...
+ * with a pivot point at ([px], [py]). The pivot point is the
+ * coordinate that should remain unchanged by the specified transformation.
+ *
+ * @param kx  The skew factor of x.
+ * @param ky  The skew factor of y.
+ * @param px  The x coordinate of the pivot point.
+ * @param py  The y coordinate of the pivot point.
+ * @return the skewed bitmap
+ */
+inline fun Bitmap.skew(kx:Float, ky:Float, px:Float = 0f, py:Float = 0f):Bitmap
```
romainguy: Remove, this is not useful

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -111,3 +116,56 @@ inline fun createBitmap(
...
+ * Creates a new bitmap, rotated from this bitmap by [degrees] - the specified number of degrees,
+ * with a pivot point at ([px], [py]). The pivot point is the coordinate that should remain
+ * unchanged by the specified transformation.
+ *
+ * @param degrees The number of degrees.
+ * @param px      The x coordinate of the pivot point.
+ * @param py      The y coordinate of the pivot point.
+ * @return the rotated bitmap
+ */
+inline fun Bitmap.rotate(degrees:Float, px: Float, py: Float)
```
romainguy: Not sure this is usefuul

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -111,3 +116,56 @@ inline fun createBitmap(
...
+ * [width],[height] are the same as the current width and height of this bitmap,
+ * this bitmap is returned and no new bitmap is created.
+ *
+ * @param x      The x coordinate of the first pixel.
+ * @param y      The y coordinate of the first pixel.
+ * @param width  The width.
+ * @param height The height.
+ * @return the clipped bitmap
+ */
+inline fun Bitmap.clip(x:Int, y:Int, width: Int, height: Int)
```
romainguy: You should run klint, your formatting is inconsistent (`x:Int` vs `with: Int` for instance)
I would rather call this method `crop()`

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -111,3 +116,56 @@ inline fun createBitmap(
...
+
+/**
+ * Returns ByteArray compressed from this bitmap with the specified [format]
+ * and [quality].
+ *
+ * @param format The format of bitmap.
+ * @param quality Hint to the compressor, 0-100. 0 meaning compress for small size, 100 meaning compress for max quality.
+ * @return ByteArray
+ */
+inline fun Bitmap.toByteArray(format:CompressFormat = CompressFormat.JPEG, @IntRange(from = 0, to = 100) quality: Int=100)
```
romainguy: I still don't find compressing to a byte array useful. Compressing directly into a stream seems the more common case

api/current.txt
feiyanke: sorry, it's a mistake, I have already removed it

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -118,17 +119,18 @@ inline fun createBitmap(
...
+ * Returns ByteArrayInputStream compressed from this bitmap with the specified [format]
  * and [quality].
  *
  * @param format The format of bitmap.
  * @param quality Hint to the compressor, 0-100. 0 meaning compress for small size, 100 meaning compress for max quality.
- * @return ByteArray
+ * @return ByteArrayInputStream
  */
-inline fun Bitmap.toByteArray(
+inline fun Bitmap.toStream(
```
romainguy: That is not what I meant. What I meant is that there shouldn't be an extension on `Bitmap` to compress to a byte array or a stream, etc. `Bitmap.compress()` already accepts an `OuputStream` which is generic enough to cover all use cases. It shouldn't be specialized here.

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -141,23 +143,9 @@ inline fun Bitmap.toByteArray(
  * @param height The height.
  * @return the clipped bitmap
  */
-inline fun Bitmap.clip(x: Int, y: Int, width: Int, height: Int) =
+inline fun Bitmap.clip(x: Int, y: Int, width: Int, height: Int): Bitmap =
```
romainguy: The return type is not needed. Please rename this method to `crop()`.

src/main/java/androidx/graphics/Bitmap.kt
```diff
@@ -111,3 +113,30 @@ inline fun createBitmap(
...
+ * @param width The width.
+ * @param height The height.
+ * @return the cropped bitmap
+ */
+inline fun Bitmap.crop(x: Int, y: Int, width: Int, height: Int) =
+    Bitmap.createBitmap(this, x, y, width, height)
+
+/**
+ * Creates a new bitmap, rotated from this bitmap by [degrees] - the specified number of degrees,
+ * with a pivot point at ([px], [py]). The pivot point is the coordinate that should remain
```
romainguy: There's no pivot point in the API

feiyanke: sorry for that, i added it.

romainguy: Thanks for the change. The pivot point should probably be set to width/2.0f and height/2.0f by default.

romainguy:
I am not against those two extensions, they are actually kind of useful, but they don't really leverage any feature of Kotlin, beyond being simple extensions. rotate with a default pivot point would.

src/main/java/androidx/graphics/Bitmap.kt
feiyanke: I added the default pivot point.

PR #419, opened 2018-03-09
Always specify return type

hendraanggrian:
Modify all code to specify return type that is not `Unit`. In some cases, doing so will provide null-safety. In most cases, it is easier to read. I strongly believe this is the way to go with all public APIs.

Also use assignment in functions whenever possible.

### DIFF
--- a/src/main/java/androidx/animation/Animator.kt
+++ b/src/main/java/androidx/animation/Animator.kt
@@ -17,6 +17,8 @@
 package androidx.animation
 
 import android.animation.Animator
+import android.animation.Animator.AnimatorListener
+import android.animation.Animator.AnimatorPauseListener
 import android.support.annotation.RequiresApi
 
 /**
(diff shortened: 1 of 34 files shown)


Sathawale27:
For some inline functions in Uri.kt return type is defined but for others it is omitted. Also see this:
https://discuss.kotlinlang.org/t/type-inference-for-return-types-of-longer-functions/554

hendraanggrian:
Hi @Sathawale27, I don't quite get what you're trying to say. I understand some return types are omitted, my point is they shouldn't because they are public APIs. And your link seems to strengthen my position since one JetBrains member explicitly said return type must be specified if **function is public or protected, i. e. it is part of API, and you don't want it's signature to change implicitly;**.

JakeWharton:
The API signature is tracked with `api/current.txt` so it can never change accidentally. You've also modified functions with bodies to expression functions in this PR which makes the diff larger than it needs to be. And finally, there are cases where we explicitly want to propagate platform types which is only possible when the return type is not explicitly specified that you've broken (hence the failing build).

We don't seem to gain anything by doing this.

Sathawale27:
@hendraanggrian When we create any extension function, It just creates the public static method and passes its reference (Object on which method is calling) as a first argument to its method and when we call it then, It just imports the package of extension function. 
@JakeWharton If it is correct then, Is it is really needed to use return type for other extension functions? ex. Uri.kt , Cursor.kt etc.

JakeWharton:
Usually to avoid platform types when invoking a factory function and not a constructor, yes.

PR #421, opened 2018-03-09
Add Path.fromPoints method.

yanglinzhen:
Add Path.fromPoints method.

### DIFF
--- a/src/androidTest/java/androidx/graphics/PathTest.kt
+++ b/src/androidTest/java/androidx/graphics/PathTest.kt
@@ -153,23 +153,22 @@ class PathTest {
         assertEquals(RectF(0.0f, 0.0f, 15.0f, 15.0f), r)
     }
 
-    @SdkSuppress(minSdkVersion = 19)
-    @Test fun testFromPoints() {
+    @Test fun testPathFromPoints() {
         val rect = RectF(0f, 0f, 40f, 40f)
 
-        val path = Path()
         val points = ArrayList<PointF>()
         points.add(PointF(rect.left, rect.top))
         points.add(PointF(rect.right, rect.top))
         try {
-            path.fromPoints(points)
+            val illegalPath = createPathFromPoints(points)
+            assertEquals(true, illegalPath.isConvex)
         } catch (e: IllegalArgumentException) {
             e.printStackTrace()
         }
         points.add(PointF(rect.right, rect.bottom))
         points.add(PointF(rect.left, rect.bottom))
-        path.fromPoints(points)
+        val legalPath = createPathFromPoints(points)
 
-        assertEquals(true, path.isConvex)
+        assertEquals(true, legalPath.isConvex)
     }
 }
--- a/src/androidTest/java/androidx/graphics/PathTest.kt
+++ b/src/androidTest/java/androidx/graphics/PathTest.kt
@@ -159,16 +159,10 @@ class PathTest {
         val points = ArrayList<PointF>()
         points.add(PointF(rect.left, rect.top))
         points.add(PointF(rect.right, rect.top))
-        try {
-            val illegalPath = createPathFromPoints(points)
-            assertEquals(true, illegalPath.isConvex)
-        } catch (e: IllegalArgumentException) {
-            e.printStackTrace()
-        }
         points.add(PointF(rect.right, rect.bottom))
         points.add(PointF(rect.left, rect.bottom))
-        val legalPath = createPathFromPoints(points)
 
-        assertEquals(true, legalPath.isConvex)
+        val path = createPathFromPoints(points)
+        assertEquals(true, path.isRect(rect))
     }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/androidx/graphics/Path.kt
```diff
@@ -116,3 +116,19 @@ inline infix fun Path.xor(p: Path): Path {
         op(p, Path.Op.XOR)
     }
 }
+
+/**
+ * Generate a closed shape from [points].
+ */
+@RequiresApi(19)
+inline fun Path.fromPoints(points: List<PointF>){
```
JakeWharton: With a name like `fromPoints` I would expect this to be a factory function but it's a mutator.

src/main/java/androidx/graphics/Path.kt
```diff
@@ -116,3 +116,19 @@ inline infix fun Path.xor(p: Path): Path {
         op(p, Path.Op.XOR)
     }
 }
+
+/**
+ * Generate a closed shape from [points].
+ */
+@RequiresApi(19)
```
JakeWharton: What makes this 19+?

src/main/java/androidx/graphics/Path.kt
```diff
@@ -116,3 +116,19 @@ inline infix fun Path.xor(p: Path): Path {
...
+inline fun Path.fromPoints(points: List<PointF>) {
```
JakeWharton: With a name like `fromPoints` I would expect this to be a factory function but it's a mutator.

src/main/java/androidx/graphics/Path.kt
yanglinzhen: Copied from method above, I'll remove it.

src/main/java/androidx/graphics/Path.kt
```diff
@@ -116,3 +116,18 @@ inline infix fun Path.xor(p: Path): Path {
         op(p, Path.Op.XOR)
     }
 }
+
+/**
+ * Create a path with closed shape from [points].
+ */
+inline fun createPathFromPoints(points: List<PointF>): Path {
+    return Path().apply {
+        if (points.size < 3) throw IllegalArgumentException("points' size should be over 2")
```
romainguy: Why should it have more than 2?

src/main/java/androidx/graphics/Path.kt
```diff
@@ -116,3 +116,18 @@ inline infix fun Path.xor(p: Path): Path {
         op(p, Path.Op.XOR)
     }
 }
+
+/**
+ * Create a path with closed shape from [points].
+ */
+inline fun createPathFromPoints(points: List<PointF>): Path {
```
romainguy: This method only creates lines, this is of rather limited use especially for a Path. Canvas has a drawLines() method for instance that can take an array of x/y coordinates to draw the path.

src/androidTest/java/androidx/graphics/PathTest.kt
```diff
@@ -152,4 +152,23 @@ class PathTest {
...
     }
+
+    @Test fun testPathFromPoints() {
+        val rect = RectF(0f, 0f, 40f, 40f)
+
+        val points = ArrayList<PointF>()
+        points.add(PointF(rect.left, rect.top))
+        points.add(PointF(rect.right, rect.top))
+        try {
+            val illegalPath = createPathFromPoints(points)
```
romainguy: I don't understand what makes such a path "illegal"

src/androidTest/java/androidx/graphics/PathTest.kt
```diff
@@ -152,4 +152,23 @@ class PathTest {
...
+            assertEquals(true, illegalPath.isConvex)
+        } catch (e: IllegalArgumentException) {
+            e.printStackTrace()
+        }
+        points.add(PointF(rect.right, rect.bottom))
+        points.add(PointF(rect.left, rect.bottom))
+        val legalPath = createPathFromPoints(points)
+
+        assertEquals(true, legalPath.isConvex)
```
romainguy: Why such emphasis on the path being convex?

src/main/java/androidx/graphics/Path.kt
romainguy: It doesn't sound efficient to first build a list of point to then create a path, when you can just create the path in the first place.

src/androidTest/java/androidx/graphics/PathTest.kt
yanglinzhen: At first I thought it should draw a contour by canvas, not a line. This is not necessarily illegal

src/main/java/androidx/graphics/Path.kt
yanglinzhen: drawLines() does not draw a closed shape

src/androidTest/java/androidx/graphics/PathTest.kt
yanglinzhen: It was not quite strict, I have changed it.

src/main/java/androidx/graphics/Path.kt
yanglinzhen: for the case I have a list of point already, and want to build a path.

romainguy: All you have to do is add an extra line at the end (it's all that Path does).

PR #427, opened 2018-03-11
Add round extensions for View

Tougee:


src/main/java/androidx/core/view/View.kt
```diff
@@ -286,3 +288,26 @@ inline fun <reified T : ViewGroup.LayoutParams> View.updateLayoutParams(block: T
...
+    clipToOutline = true
+}
+
+/**
+ * Sets the Outline to the view defined by the corner radius.
+ *
+ * @see Outline.setRoundRect
+ */
+@RequiresApi(21)
+inline fun View.round(@Px radius: Int) = round(radius.toFloat())
```
jaredsburrows: 👀

romainguy: I don't see why this method is needed.

src/main/java/androidx/core/view/View.kt
```diff
@@ -286,3 +288,26 @@ inline fun <reified T : ViewGroup.LayoutParams> View.updateLayoutParams(block: T
...
     layoutParams = params
 }
+
+/**
+ * Sets the Outline to the view defined by the corner radius.
+ *
+ * @see Outline.setRoundRect
+ */
+@RequiresApi(21)
+inline  fun View.round(@Px radius: Float) {
```
romainguy: This seems a little too specific. The name of the method isn't great either.

Tougee: Ok I'll close it

LOH9999:
Ok

PR #430, opened 2018-03-13
Cursor extension

feiyanke:
add forEach() method to iterate Cursor


JakeWharton:
See #201 for a discussion about why this isn't included.

PR #434, opened 2018-03-13
Add context extensions for getBoolean, getInteger, getDimension.

CosminMihuMDC:
Add Context extensions for Resources methods:

getBoolean -> https://developer.android.com/reference/android/content/res/Resources.html#getBoolean(int)
getInteger -> https://developer.android.com/reference/android/content/res/Resources.html#getInteger(int)
getDimensions -> https://developer.android.com/reference/android/content/res/Resources.html#getDimension(int)


JakeWharton:
These are resources. Why should we put them on context?

CosminMihuMDC:
@JakeWharton 

We already have Context.getString by default in Android Framework. (see https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/java/android/content/Context.java#LC555)
So, it could be a good idea to have also other simple methods from Resources in Context to be easily accessed like getString.

JakeWharton:
Then I think we should either deprecate that method or add these to Context in the framework.

romainguy:
`getString()` exists on `Context` because it's a very common task to do, a lot more than `getBoolean()` for instance.

CosminMihuMDC:
@romainguy You're right. `getBoolean` and `getInteger` are not so often used. On the other hand,  `getDimension` could be really useful as `getString` is.

JakeWharton:
But there's two useful variants of retrieving dimensions and you've (arguably) included the less-useful variant despite it's more appealing name.

PR #436, opened 2018-03-14
SpannableStringBuilder extensions

feiyanke:
add wrappers for SuperscriptSpan, SubscriptSpan, TypefaceSpan and URLSpan

### DIFF
--- a/src/main/java/androidx/text/SpannableStringBuilder.kt
+++ b/src/main/java/androidx/text/SpannableStringBuilder.kt
@@ -27,6 +27,10 @@ import android.text.style.ForegroundColorSpan
 import android.text.style.StrikethroughSpan
 import android.text.style.RelativeSizeSpan
 import android.text.style.StyleSpan
+import android.text.style.SubscriptSpan
+import android.text.style.SuperscriptSpan
+import android.text.style.TypefaceSpan
+import android.text.style.URLSpan
 import android.text.style.UnderlineSpan
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/androidx/text/SpannableStringBuilder.kt
```diff
@@ -134,3 +138,39 @@ inline fun SpannableStringBuilder.scale(
...
+ */
+inline fun SpannableStringBuilder.up(builderAction: SpannableStringBuilder.() -> Unit) =
+    inSpans(SuperscriptSpan(), builderAction = builderAction)
+
+/**
+ * Wrap appended text in [builderAction] in a [SubscriptSpan].
+ *
+ * @see SpannableStringBuilder.inSpans
+ */
+inline fun SpannableStringBuilder.down(builderAction: SpannableStringBuilder.() -> Unit) =
```
JakeWharton: I don't think "up" and "down" correctly convey the behavior. "superscript" and "subscript" seem fine.

Also are these common enough that they warrant inclusion? I can't remember using them once.

src/main/java/androidx/text/SpannableStringBuilder.kt
```diff
@@ -134,3 +138,39 @@ inline fun SpannableStringBuilder.scale(
...
+    inSpans(SubscriptSpan(), builderAction = builderAction)
+
+/**
+ * Wrap appended text in [builderAction] in a [TypefaceSpan].
+ *
+ * @see SpannableStringBuilder.inSpans
+ */
+inline fun SpannableStringBuilder.font(
```
JakeWharton: I'm torn between calling this "font" or "typeface" to match the span name. I suspect matching the span name will improve discoverability.

feiyanke: Matching name seems like better, I renamed the methods - "up", "down", "font" to "superscript", "subscript" and "typeface" .

Issue #265, opened 2018-02-07
API for getting intent extras within an Activity implementation


gabrielhuff:
I think it would be nice to have something for accessing argument intent extras within an `Activity` implementation.

I'm proposing two alternatives here (I'd personally go for the second one):

### Alternative 1 - Minimalist, unsafe at compile time

```kotlin
class MyActivity : Activity {
  
  private val id: Long? by extra("id")

  private val timestamp: Long by extra("timestamp")

  private val name: String by extra("name")
}

// Pseudo implementation, just to show the equivalency to lazy
inline fun <reified T> Activity.extra(key: String): ReadOnlyProperty<Activity, T> {
  return lazy { intent.get(key) as T }
}
```

### Alternative 2 - Robust, safe at compile time

```kotlin
class MyActivity : Activity {
  
  private val id by Extra.Nullable.long("id")

  private val timestamp by Extra.long("timestamp")

  private val name by Extra.string("name")
}

object Extra {

    fun long(key: String): ReadOnlyProperty<C, Long> = TODO()

    fun string(key: String): ReadOnlyProperty<C, Int> = TODO()

    // ...
    
    object Nullable {

        fun long(key: String): ReadOnlyProperty<C, Long> = TODO()

        // ...
    }
}
```

dovahkiin98:
I don't think they're a method "intent.get(key)", intent is very specific about the return type

```kotlin
inline fin <T> Activity.intentExtra(key: String, defaultValue: T) = lazy {
    when(T::class.java) {
        Int::class.java -> intent.getIntExtra(key, defaultValue) as T
        Float::class.java -> intent.getFloatExtra(key, defaultValue) as T
...
    }
}

//call:
val intExtra: Int by intentExtra("intExtra", 0)
```

I don't think it would be any better if we make it like this
```kotlin
fun Intent.get(key)

//call
intent["key"]?: defaultValue
```

PR #440, opened 2018-03-17
Property delegates for Activity extras

gabrielhuff:
This implements #265. Usage:

```kotlin
class MyActivity : Activity {
  
  private val id: Long? by extra("id")

  private val timestamp: Long by extra("timestamp")

  private val name: String by extra("name")
}
```

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -1,3 +1,11 @@
+package androidx.core.activity {
+
+  public final class ActivityKt {
+    ctor public ActivityKt();
+  }
+
+}
+
 package androidx.core.animation {
 
   public final class AnimatorKt {
--- a/src/androidTest/AndroidManifest.xml
+++ b/src/androidTest/AndroidManifest.xml
@@ -3,5 +3,6 @@
     <application>
         <activity android:name="androidx.core.TestActivity"/>
         <activity android:name="androidx.core.TestPreferenceActivity"/>
+        <activity android:name="androidx.core.TextExtraActivity"/>
     </application>
 </manifest>
--- a/src/androidTest/java/androidx/core/TextExtraActivity.kt
+++ b/src/androidTest/java/androidx/core/TextExtraActivity.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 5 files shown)


gabrielhuff:
I signed it!

dovahkiin98:
You're missing the default value parameter

gabrielhuff:
@dovahkiin98 What do you mean? Where?

dovahkiin98:
@gabrielhuff when you use intent.getStringExtra("extra", "") you provide a default value, it's not always 0 or "" or false, sometimes it's different.
I would suggest this implementation :  
```kotlin
inline fun <T> Activity.extra(key: String, defaultValue: T) = lazy {
    when(T::class.java) {
        Int::class.java -> intent.getIntExtra(key, defaultValue) as T
        Float::class.java -> intent.getFloatExtra(key, defaultValue) as T
...
    }
}

val intExtra: Int by extra("intExtra", 2)
```

gabrielhuff:
@dovahkiin98 The current implementation allows you to define default values. The only difference is that we are getting them from a function instead of a raw value:

```kotlin
val intExtra by extra("intExtra") { 2 }
```

Note that you can omit the default value function. If so, the implementation will throw an `IllegalArgumentException` if no extra was passed.

JakeWharton:
Delegates are an expensive hammer to use for this. I don't think we've decided whether their use is appropriate for simple things like this.

gabrielhuff:
IMHO delegates (in this case, specifically lazy instantiation) are perfect for this. Consider a scenario in which you need to access the same extra on different scopes. What would be the most appropriate way to do it?

JakeWharton:
A `lateinit` property is an order of magnitude less expensive.

I have no idea what "same extra on different scopes" means.

gabrielhuff:
By different scopes I mean different methods / contexts. We don't want to read the same extra from the input intent more than once (not because of performance - just because the call might be kinda ugly):

```kotlin
override fun onCreate(state: Bundle?) {
  ...
  val extra = intent.getParcelableExtra("my_key") as MyCustomExtra
}

fun onSomethingElse() {
  val extra = intent.getParcelableExtra("my_key") as MyCustomExtra // Eks
}
```

But I get what you're saying. Performance-wise, `lateinit` would be ideal. However, the property would have to be a `var` and would require an additional line of code somewhere else for instantiation.

PR #442, opened 2018-03-19
Get the MIME type from File

tobiasschuerg:
Get the MIME type of a `File` via 
```kotlin
val type = file.getMimeType()
```

### DIFF
--- a/src/main/java/androidx/core/io/File.kt
+++ b/src/main/java/androidx/core/io/File.kt
@@ -1,0 +1,31 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.core.io
+
+import android.webkit.MimeTypeMap
+import java.io.File
+
+/**
+ * Returns the MIME type (content type) of this [File].
+ *
+ * @see MimeUtils
+ * @return The MIME type for the files' extension or null iff there is none.
+ */
+fun File.getMimeType(): String? {
+    return MimeTypeMap.getFileExtensionFromUrl(toString())
+        ?.apply { MimeTypeMap.getSingleton().getMimeTypeFromExtension(toLowerCase()) }
+}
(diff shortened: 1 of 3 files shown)


Sathawale27:
@tobiasschuerg What if there are spaces in your file path? It will not work in that condition. You need use encode uri for that purpose.

Like,

```
val file = File("mnt/sdcard/abc test file.png")
```

src/main/java/androidx/core/io/File.kt
```diff
@@ -0,0 +1,32 @@
...
+import java.io.File
+
+/**
+ * Returns the MIME type (content type) of this [File].
+ *
+ * @see MimeUtils
+ * @return The MIME type for the files' extension or null iff there is none.
+ */
+fun File.getMimeType(): String? {
+    return MimeTypeMap.getFileExtensionFromUrl(toString().urlEncode())
```
Sathawale27: IMHO, We can use path instead of toString as it easy to understand.

src/main/java/androidx/core/text/String.kt
```diff
@@ -14,10 +14,13 @@
  * limitations under the License.
  */
 
-@file:Suppress("NOTHING_TO_INLINE") // Aliases to public API.
+@file:Suppress("NOTHING_TO_INLINE")
+
+// Aliases to public API.
```
Sathawale27: nit: Remove extra spaces. Keep this consistent as others.

src/main/java/androidx/core/io/File.kt
```diff
@@ -0,0 +1,32 @@
...
+
+import android.webkit.MimeTypeMap
+import androidx.core.text.urlEncode
+import java.io.File
+
+/**
+ * Returns the MIME type (content type) of this [File].
+ *
+ * @see MimeUtils
+ * @return The MIME type for the files' extension or null iff there is none.
```
Sathawale27: nit: There should be "if" instead of "iff"

src/main/java/androidx/core/io/File.kt
```diff
@@ -0,0 +1,32 @@
...
+import androidx.core.text.urlEncode
+import java.io.File
+
+/**
+ * Returns the MIME type (content type) of this [File].
+ *
+ * @see MimeUtils
+ * @return The MIME type for the files' extension or null if there is none.
+ */
+fun File.getMimeType(): String? {
```
mekarthedev: The syntax seems confusing:

1. In Kotlin we have properties. Declaring a method instead of a property may be used as a way of paying reader's attention to a possible heavy computations behind the call. I'm not sure about `getMimeTypeFromExtension`'s performance. But if it is not so time consuming, declaring it as a property seems a better choice.
2. The name. After I've read `file.getMimeType()` the first time, my first impression was "hmm, does it guesses mime type by reading first few bytes of a file?". Then I read implementation and found it actually uses `getMimeTypeFromExtension` and doesn't read file content. Note how SDK's naming pays attention to the way the mime type is guessed. This implementation detail may be important at the call site.

I'd suggest to declare it as `val File.mimeTypeFromExtension: String?` or `fun File.guessMimeTypeFromExtension(): String?`.

PR #444, opened 2018-03-21
Add context extension for making call.

yuanhoujun:
Simplify making call api

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -23,6 +23,7 @@ package androidx.core.content {
 
   public final class ContextKt {
     ctor public ContextKt();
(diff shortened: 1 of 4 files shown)


JakeWharton:
Thanks for the PR, but this is an _extremely_ specific utility method that's not appropriate for this library.

yuanhoujun:
So what's the appropriate api , what is the difference between this library and Anko?

JakeWharton:
This library is extensions for adapting the platform APIs to be idiomatic in Kotlin by leveraging features like extensions, reified generics, and lambdas with receivers.

PR #445, opened 2018-03-21
Add context extension for getting status bar height.

yuanhoujun:
Maybe this is needed.

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -23,6 +23,7 @@ package androidx.core.content {
 
   public final class ContextKt {
     ctor public ContextKt();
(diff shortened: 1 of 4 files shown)


JakeWharton:
This resource ID isn't public and cannot be relied on. The correct way to position your content below system UI like the status bar or navigation bar is with window insets. There may be anywhere from 0 to 4 insets on the 4 sides of the screen and depending on orientation, and with Android P there also is the availability of a display notch.

More info:
* https://medium.com/google-developers/why-would-i-want-to-fitssystemwindows-4e26d9ce1eec
* https://www.youtube.com/watch?v=_mGDMVRO3iE
* https://developer.android.com/preview/features.html#cutout

PR #464, opened 2018-03-27
Add base 64 encoding/decoding extensions.

JakeWharton:
I'm really not sure what to do about the parameter order. `flags` seem like the most useful and they're present on every overload so I put them first, but it's definitely weird to have `offset`/`length` be not first. Thoughts?

Closes #406.

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -450,6 +450,14 @@ package androidx.core.util {
     method @RequiresApi(17) public static void writeText(android.util.AtomicFile, String text, java.nio.charset.Charset charset = "Charsets.UTF_8");
   }
 
+  public final class Base64Kt {
+    ctor public Base64Kt();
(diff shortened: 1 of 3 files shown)


jaredsburrows:
I have signed the @google CLA(check previous PRs) and I am in the @google organization: https://github.com/jaredsburrows.

jaredsburrows:
Ideally it would nice to keep parameters in order with the method but `flags` does seem to be more favorable as the first option.

Sathawale27:
In most of the cases we need to encode/decode full ByteArray to/from Base64. So, length and offset seems perfectly fine after flags in order. I have one question in its implementation that, Can't we use Base64.encodeToString function which is already available in android API instead of encoding a ByteArray to byte[] and then toString with ASCII.

JakeWharton:
Yep!

jaredsburrows:
@romainguy @JakeWharton Review :)

PR #465, opened 2018-03-27
Add View.setDebounceClickListener() extension

TinoGuo:
useful extension for preventing user click too fast

### DIFF
--- a/src/main/java/androidx/core/view/View.kt
+++ b/src/main/java/androidx/core/view/View.kt
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
-@file:Suppress("NOTHING_TO_INLINE") // Aliases to other public API.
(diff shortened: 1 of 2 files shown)


src/main/java/androidx/core/view/View.kt
```diff
@@ -286,3 +290,26 @@ inline fun <reified T : ViewGroup.LayoutParams> View.updateLayoutParams(block: T
...
     layoutParams = params
 }
+
+/**
+ * Executes the block only when the [duration] longer than the duration of two click timestamp
+ * default debounce time is 1 second
+ * @param duration [Long]
+ * @param timeUnit [TimeUnit]
+ */
+fun View.setDebounceClickListener(
```
romainguy: android-ktx's goal is not to add new features. Such a feature should be submitted to AOSP.

PR #466, opened 2018-03-27
Add View.scale extension

armcha:
Added View.scale extension

### DIFF
--- a/src/main/java/androidx/core/view/View.kt
+++ b/src/main/java/androidx/core/view/View.kt
@@ -286,3 +286,18 @@ inline fun <reified T : ViewGroup.LayoutParams> View.updateLayoutParams(block: T
     block(params)
     layoutParams = params
 }
+
+/**
+ * Setting both x and y scale to view
+ *
+ * ```
+ * view.scale = 1.2f
+ * ```
+ * Returns the minimum scale of scaleY and scaleX
+ */
+var View.scale: Float
+    get() = Math.min(scaleX, scaleY)
+    set(value) {
+        scaleY = value
+        scaleX = value
+    }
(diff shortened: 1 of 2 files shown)


armcha:
@googlebot I signed it!

src/main/java/androidx/core/view/View.kt
```diff
@@ -286,3 +286,18 @@ inline fun <reified T : ViewGroup.LayoutParams> View.updateLayoutParams(block: T
...
+/**
+ * Setting both x and y scale to view
+ *
+ * ```
+ * view.scale = 1.2f
+ * ```
+ * Returns the minimum scale of scaleY and scaleX
+ */
+var View.scale: Float
+    get() = Math.min(scaleX, scaleY)
```
romainguy: This is bizarre.

src/main/java/androidx/core/view/View.kt
```diff
@@ -286,3 +286,18 @@ inline fun <reified T : ViewGroup.LayoutParams> View.updateLayoutParams(block: T
...
+
+/**
+ * Setting both x and y scale to view
+ *
+ * ```
+ * view.scale = 1.2f
+ * ```
+ * Returns the minimum scale of scaleY and scaleX
+ */
+var View.scale: Float
```
romainguy: I'm not convinced this is a common thing to do on View. Animations are the main use case for scale and this doesn't help animations.

armcha: @romainguy ok, thanks. Then I thing ViewPropertyAnimator.scale is a good idea.

JakeWharton: Please file a feature request for that at http://b.android.com/new

PR #475, opened 2018-03-30
Improve header comment  in Size.kt

toshihirooya:
I changed header comment from 20188 to 2018 in Size.kt.

That’s because, I guess 20188 in header comment mean generated-date of this class.
So, it should be change.


romainguy:
Thank you!

toshihirooya:
Thanks!!!

Issue #476, opened 2018-03-30
SharedPreferences putDouble


Bloody-Badboy:
SharedPreferences Extension function to set a double value in the preferences and to retrieve the double value from the preferences

```kotlin
inline fun SharedPreferences.Editor.putDouble(key: String, value: Double): SharedPreferences.Editor {
    putLong(key, value.toRawBits())
    return this
}

inline fun SharedPreferences.getDouble(key: String, defValue: Double) = Double.fromBits(getLong(key, defValue.toRawBits()))
```

JakeWharton:
If this has merit then it seems like something we should just add to the platform. There's nothing special about Kotlin code that makes it such that they should be the only ones that can easily read/write doubles from shared preferences.

JakeWharton:
Please file a feature request for this to be added to the framework at http://b.android.com/new

PR #479, opened 2018-03-30
SharedPreferences.Editor.putDouble() and SharedPreferences.getDouble()

Bloody-Badboy:
Add SharedPreferences.Editor.putDouble() and SharedPreferences.getDouble()

```kotlin
inline fun SharedPreferences.Editor.putDouble(
    key: String,
    value: Double
): SharedPreferences.Editor {
    putLong(key, value.toRawBits())
    return this
}
```
```kotlin
inline fun SharedPreferences.getDouble(key: String, defValue: Double): Double {
    return Double.fromBits(getLong(key, defValue.toRawBits()))
}
```


JakeWharton:
Can you address https://github.com/android/android-ktx/issues/476#issuecomment-377521719?

PR #483, opened 2018-04-02
Add View.setUsageHint() to describe click actions for TalkBack users

ataulm:
Usage hints are used by TalkBack to describe the action that will be triggered on click or long click, e.g.:

`"The Matrix... <pause>... Double tap to start playing movie"`

where `"The Matrix"` is the content description and `"Double tap to start playing movie"` is the usage hint. 

Without a _custom_ hint set, it will read: `"The Matrix... <pause>... Double tap to activate"`.

---

**Java implementation:**

```java
    ...
    setUsageHint(movieItemView, "start playing movie", null);
}

private static void setUsageHint(View view, @Nullable CharSequence clickLabel, @Nullable CharSequence longClickLabel) {
    ViewCompat.setAccessibilityDelegate(
        this, new AccessibilityDelegateCompat() {
            @Override
            public void onInitializeAccessibilityNodeInfo(View host, AccessibilityNodeInfoCompat info) {
                super.onInitializeAccessibilityNodeInfo(host, info);

                if (host.isClickable()) {
                    info.addAction(new AccessibilityNodeInfoCompat.AccessibilityActionCompat(ACTION_CLICK, clickLabel));
                }

                if (host.isLongClickable()) {
                    info.addAction(new AccessibilityNodeInfoCompat.AccessibilityActionCompat(ACTION_LONG_CLICK, longClickLabel));
                }
            }
        }
    );
}
```

**with Kotlin-ktx**:

```
movieItemView.setUsageHint("start playing movie")
```
---

There are many more combinations of tests I can add but I dunno if it's too much. I am concerned that `View.kt` and consequently `ViewTest.kt` will get very large - is there some strategy to mitigate this or is it not considered an issue?

---

On pre-Lollipop, this behaviour is a no-op (handled by `ViewCompat`'s implementation of `AccessibilityDelegate` under the hood).

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -637,6 +637,8 @@ package androidx.core.view {
     method public static void setGone(android.view.View, boolean value);
     method public static void setInvisible(android.view.View, boolean value);
     method public static void setPadding(android.view.View, @Px int size);
+    method public static void setUsageHint(android.view.View, @StringRes Integer? clickLabel = "null", @StringRes Integer? longClickLabel = "null");
+    method public static void setUsageHint(android.view.View, CharSequence? clickLabel = "null", CharSequence? longClickLabel = "null");
     method public static void setVisible(android.view.View, boolean value);
     method public static android.graphics.Bitmap toBitmap(android.view.View, android.graphics.Bitmap.Config config = "Bitmap.Config.ARGB_8888");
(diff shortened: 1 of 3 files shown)


ataulm:
⚠️ Note: `ViewCompat` comes from the support library.

This functionality is possible without `ViewCompat`, since `View` also has `setAccessibilityDelegate` but would require `RequiresSdk` annotations. I don't mind either.

ataulm:
Is there something I could answer or change to get this merged in?

If it's not suitable for this library, could we close the PR?

PR #489, opened 2018-04-09
Add extensions for Uri

feiyanke:
Add extensions of Uri for ContentUris


JakeWharton:
This looks good. I never knew this class existed. I need to double check it once I'm not on a phone.

PR #502, opened 2018-04-16
Recycle TypedArray instance even if use block threw an exception.

aleksandrsorlovs:
Current version of `TypedArray.use` extension function doesn't call `TypedArray.recycle` if exception was thrown. 
`Closeable.use` (which is referenced in a javadoc to `TypedArray.use`) calls `Closeable.close` in all cases, and if exception was thrown during closing it is also added to the primary exception.
New `TypedArray.use` implementation is consistent with `Closeable.use`.


JakeWharton:
TypedArray instances are pooled and not a resource. There's no need to add the overhead of a try/catch to something that is both extremely unlikely and also should be taking the whole process down.

PR #506, opened 2018-04-17
Add Collection extension (runIfNotEmpty)

wlals822:
I created an extension to handle empty collection.

### Usage 
```kotlin
val someCollection = arrayListOf<String>("foo", "bar")
someCollection.runIfNotEmpty {
     forEach {
        print(it)
     }
} 

val emptyCollection = Collections.emptyList<Any>()

emptyCollection.runIfNotEmpty {
       // skips this block
}
```

### DIFF
--- a/src/androidTest/java/androidx/core/util/CollectionsTest.kt
+++ b/src/androidTest/java/androidx/core/util/CollectionsTest.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2017 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


simonschiller:
I don't see how 

    collection.runIfNotEmpty {
        // Some action
    }

is better than

    if (collection.isNotEmpty()) {
        // Some action
    }

JakeWharton:
This library covers types in the Android package only. Please suggest extensions for types in the java package to the Kotlin standard library.

wlals822:
@simonschiller I intended to use `this` in the { } block . But yes, it's not big deal except for one less variable. thx for the review!
@JakeWharton I see. I will consider something like that.

Issue #497, opened 2018-04-15
Extension function to work with ClipData


AkshayChordiya:
`ClipData` holds multiple URIs and it is generally useful if you are firing a chooser intent with ability to select multiple items. In those cases, the intent returned contains `ClipData` which in turn provides list of URIs of the selected items.

Previously we needed to write the boilerplate code to get list of URIs from `ClipData` and vice versa. Instead I thought to utilize the extension functions to make it simple and clean.

Below are my proposed functions to convert `ClipData` to list of URIs and the reverse from list of URIs to `ClipData`:

**ClipData to List**
```
fun ClipData.toList(): MutableList<Uri> {
    val length = itemCount
    val result = mutableListOf<Uri>()
    for (i in 0 until length) {
        result.add(getItemAt(i).uri)
    }
    return result
}
```

**List to ClipData**
```
fun List<Uri>.toClipData(label: String): ClipData {
    val clipData = ClipData(
            ClipDescription(label, arrayOf(ClipDescription.MIMETYPE_TEXT_URILIST)),
            ClipData.Item(get(0))
    )
    for (i in 1..size) {
        clipData.addItem(ClipData.Item(get(i)))
    }
    return clipData
}
```

I'm open to suggestions, btw.

jzoran:
Since there are other interesting properties of ClipData.Item, e.g. you're working with clipped text or shortcuts (that use intents if I'm not mistaken) etc., why not something along the lines of: 
```
inline fun <reified T> ClipData.map(transform:  (ClipData.Item) -> T): List<T> {
    val n = this.itemCount
    var m = mutableListOf<T>()
    for (i in 0 until n) {
        m.add(transform(this.getItemAt(i)))
    }
    return m.toList()
}
```

Then, of course on caller's side your use-case would look like:
```
var c: ClipData
...
val l = c.map { item -> item.uri }
```
Also, ClipData could use an iterator and a forEach extension (similar to Menu's extensions), in which case the above definition would be slightly nicer on the eyes:
```
inline fun <reified T> ClipData.map(transform:  (ClipData.Item) -> T): List<T> {
    var m = mutableListOf<T>()
    this.foreach { 
        m.add(transform(it))
    }
    return m.toList()
}
```
(or something more appropriate).

I'd be happy to make a PR.

AkshayChordiya:
This looks more interesting and generic. That's the reason why I didn't PR in first place.
Let's wait for more suggestions or feedback and then head for PR

[...]

PR #508, opened 2018-04-17
Add extensions for ClipData (#497)

jzoran:


JakeWharton:
Moved to r.android.com/720972

PR #509, opened 2018-04-17
Add missing Menu extensions

rom4ek:
There is a missing possibility to make operations over a `Sequence` of menu items, due to missing `children` extension, so I've added it. Also `minusAssign`.

### DIFF
--- a/src/androidTest/java/androidx/core/view/MenuTest.kt
+++ b/src/androidTest/java/androidx/core/view/MenuTest.kt
@@ -45,6 +45,20 @@ class MenuTest {
         assertTrue(item2 in menu)
     }
 
+    @Test fun minusAssign() {
+        val item1 = menu.add(NONE, 1, NONE, "")
+        val item2 = menu.add(NONE, 2, NONE, "")
+
+        assertEquals(2, menu.size)
+
+        menu -= item2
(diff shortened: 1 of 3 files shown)


romainguy:
Seems fine to me, but I can't think of a good use case for this. I'm curious about what you'd want to do with it.

rom4ek:
And also for the sake of consistency with other collection-like extensions

src/androidTest/java/androidx/core/view/MenuTest.kt
```diff
@@ -132,4 +146,17 @@ class MenuTest {
...
         assertEquals(0, menu.size())
     }
+
+    @Test fun children() {
+        val items = listOf(
+            menu.add(NONE, 1, NONE, ""),
+            menu.add(NONE, 2, NONE, ""),
+            menu.add(NONE, 3, NONE, "")
+        )
+        items.forEach { menu.add(it.groupId, it.itemId, it.order, it.title) }
```
JakeWharton: Is there any reason for doing this?

rom4ek: Oops, sorry, didn't notice the the items have already been added upon filling list up.

PR #518, opened 2018-04-21
Add proposal for LiveData extension, chaining like RxJava

henrytao-me:
Usage: 
```kotlin
val liveData: MutableLiveData<Boolean> = MutableLiveData()
liveData
    .filter { it == false }
    .map { true }
    .nonNull()
    .observe(this, { 
        // it is non-null and always true
    })
```


JakeWharton:
Thanks for the PR, but this project only covers platform APIs and does not include extensions for the support library or architecture components.

Issue #522, opened 2018-04-26
TextView TextWatcher 


zack-barakat:
Similarly to Animator.Kt can we have callback for TextView Listener
Do you think we need to simplify TextWatch implementation to pass actions instead of the whole interface. so we can have something like below.
```
textview.beforeTextChanged { s, start, count, after -> }
textview.onTextChanged { s, start, count, after -> }
textview..afterTextChanged {}
```

JakeWharton:
Text watcher-based listeners have been merged in http://r.android.com/720364

PaulWoitaschek:
And why can't I find them?

JakeWharton:
Probably because you're not using the version they are released in. Or one or a million other problems. You provided zero information so I'm not sure what kind of response you were expecting.

poldz123:
@JakeWharton How does one able to add the updated ktx with widget support dependencies to the gradle? I know it is not in the `1.0.0` since I am using it.

jnsnpg:
I can't find them too

PR #523, opened 2018-04-26
Add TextView listener extensions

zack-barakat:
This PR will add TextWatch implementation to pass actions instead of the whole interface. 
See Issue
#522

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -620,6 +620,14 @@ package androidx.core.view {
     method public static operator void minusAssign(android.view.Menu, android.view.MenuItem item);
   }
 
+  public final class TextViewKt {
+    ctor public TextViewKt();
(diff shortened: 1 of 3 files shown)


zack-barakat:
In my local environment it actually passed the tests
but something wrong with firebase cloud test:

```
#!/bin/bash -eo pipefail
ftl-tests/setup.sh
GCLOUD_SERVICE_KEY env variable is empty. Exiting.
Exited with code 1
```

chrnie233:
TextView add a TextWatcher  but never can remove, I do not think is good idea.

zack-barakat:
@chrnie233  The return type for all the extension functions is the watcher, thus you can remove it by as with the function `removeTextChangedListener(watcher)`

```
val watcher = etEmail.afterTextChanged { etEmail.error = null }
txtReason.removeTextChangedListener(l)
```

JakeWharton:
Text watcher-based listeners have been merged in http://r.android.com/720364

PR #525, opened 2018-05-02
Add two more extensions for shared preferences.

alirezaafkar:
This adds two extensions to shared preferences for put/get all types with one method.

**Examples**:

`preferences.put("user_id", 100)`

`preferences.put("user_id", 100, commit = true)`

`val userId = preferences.get("user_id", -1)`


alirezaafkar:
I signed it!

alirezaafkar:
I signed it!

PR #528, opened 2018-05-08
Add JsonObject checking value

budioktaviyan:
The idea is to simplify checking unexpected value before putting into ```JSONObject```


MGaetan89:
This library only target the Android API (in the `android`) package.

budioktaviyan:
Thanks @MGaetan89 . I will take a look to another core extensions 😄

PR #530, opened 2018-05-08
Add support for image view filtering

budioktaviyan:
This api are used to support an image view extensions in the future. At the first time, I'm adding a function to filtering image view based on its saturation

### DIFF
--- a/src/androidTest/java/androidx/core/widget/ImageViewTest.kt
+++ b/src/androidTest/java/androidx/core/widget/ImageViewTest.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright (C) 2017 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.core.widget
+
+import android.support.test.InstrumentationRegistry
+import android.support.test.filters.SdkSuppress
+import android.support.test.rule.ActivityTestRule
+import android.widget.ImageView
+import androidx.core.kotlin.test.R
+import androidx.core.TestActivity
+import org.junit.Rule
+import org.junit.Test
+
+@SdkSuppress(minSdkVersion = 16)
+class ImageViewTest {
+    @JvmField @Rule val rule = ActivityTestRule<TestActivity>(TestActivity::class.java)
+
+    @Test fun setFilter() {
+        rule.runOnUiThread {
+            val view = rule.activity.findViewById<ImageView>(R.id.image_view)
+            view.filter(0f)
+        }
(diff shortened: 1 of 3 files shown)


src/androidTest/java/androidx/core/widget/ImageViewTest.kt
```diff
@@ -0,0 +1,39 @@
...
+@SdkSuppress(minSdkVersion = 16)
+class ImageViewTest {
+    @JvmField @Rule val rule = ActivityTestRule<TestActivity>(TestActivity::class.java)
+
+    @Test fun setFilter() {
+        rule.runOnUiThread {
+            val view = rule.activity.findViewById<ImageView>(R.id.image_view)
+            view.filter(0f)
+        }
+        InstrumentationRegistry.getInstrumentation().waitForIdleSync()
```
romainguy: I don't understand what you are testing here. You wait for idle sync but you are not verifying anything.

src/main/java/androidx/core/widget/ImageView.kt
```diff
@@ -0,0 +1,33 @@
...
+import android.graphics.ColorMatrixColorFilter
+import android.graphics.Paint
+import android.widget.ImageView
+
+/**
+ * Filtering an image based on saturation
+ *
+ * @param saturation Value to set the saturation
+ */
+fun ImageView.filter(saturation: Float) {
```
romainguy: This is way too specific of an API. At least if the method was called `saturation()` it would be more sensible. But even then, it doesn't feel general purpose enough. It would be better to make color matrix/filters easier to create.

src/main/java/androidx/core/widget/ImageView.kt
```diff
@@ -0,0 +1,33 @@
...
+    val colorMatrix = ColorMatrix().apply { setSaturation(saturation) }
+    val paint = Paint().apply { colorFilter = ColorMatrixColorFilter(colorMatrix) }
+    this.setLayerType(ImageView.LAYER_TYPE_HARDWARE, paint)
```
romainguy: Why force a hardware layer? It's wasteful on an ImageView, even with a color matrix filter.

src/androidTest/java/androidx/core/widget/ImageViewTest.kt
budioktaviyan: hi, sorry i'm still not sure how to test with very small things like this :) ... but, thanks i will update the test along with some fixing there

src/main/java/androidx/core/widget/ImageView.kt
budioktaviyan: Thanks @romainguy for your input. it's true, I created this feature based on My problem when filtering an image to grey scale only that take a parameter from saturation. Good insight from you to make this more general, cool!

src/main/java/androidx/core/widget/ImageView.kt
budioktaviyan: When I'm not forcing this to hardware layer, sometimes on several device doesn't work. Do you have any suggestion ?

budioktaviyan:
@rharter I'm not sure about that, maybe @romainguy can explain us better ?

romainguy:
I wouldn't worry too much about it since it's not different from what you would do without extensions. But I don't think `ImageView` is the right place for this.

src/main/java/androidx/core/widget/ImageView.kt
romainguy: Hard to tell without knowing how you were trying to apply the filter to your image.

PR #531, opened 2018-05-08
Stared adding number functions 

anshadali:
Added function to round Double value to specified number of decimal places.


JakeWharton:
This functionality is not Android-specific and thus is not appropriate for this library. If you want behavior for numbers, please file a feature request to the Kotlin standard library.

PR #532, opened 2018-05-08
startactivity made simple

anshadali:
Now new activity will be started by calling startNewActivity(ActivityNama::class.java) from an activity


romainguy:
Thank you for your PR. This PR has several issues: you modified many files that do not need to be modified. The extension you are proposing also doesn't really add much value. All it does is save the invocation of the Intent constructor. It doesn't really take advantage of Kotlin features.

Zhuinden:
The name `startNewActivity` is rather strange, and it's not type-safe.

Just like anko, a possible and proper solution would be:

    inline fun <reified T: Activity> Context.startActivity() {
        startActivity(Intent(this, T::class.java))
    }

But whether that's needed or not is up to the KTX maintainers.

PR #534, opened 2018-05-11
Adding URI encoding and decoding extensions

RamV13:
**Before:**
```kotlin
val string = "http%3A%2F%2Ftest.example.com%2Ffoo%20bar"
val decoded = Uri.decode(string)
val encoded = Uri.encode(decoded)
```
**After:**
```kotlin
val string = "http%3A%2F%2Ftest.example.com%2Ffoo%20bar"
val decoded = string.decode()
val encoded = string.encode()
```

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -322,6 +322,8 @@ package androidx.core.net {
 
   public final class UriKt {
     ctor public UriKt();
+    method public static String decode(String);
+    method public static String encode(String, String? allow = "null");
     method public static java.io.File toFile(android.net.Uri);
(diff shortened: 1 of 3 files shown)


rharter:
This seems rather specific. From the API it's unclear what is being encoded or decoded. `string.decode()` doesn't indicate that this should decoded a URI vs. MD5 vs. SHA-1...

IMHO the `Uri.decode(string)` is a clearer API and the same amount of code so what's the added benefit of this?

RamV13:
Agreed

PR #535, opened 2018-05-11
Adding function to seamlessly create BroadcastReceiver instances

RamV13:
**Before:**
```kotlin
val mReceiver = object : BroadcastReceiver() {
    override fun onReceive(context: Context, intent: Intent) {
        // body
    }
}
```
**After:**
```kotlin
val mReceiver = broadcastReceiver { context, intent ->
    // body
}
```

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -16,6 +16,11 @@ package androidx.core.animation {
 
 package androidx.core.content {
 
+  public final class BroadcastReceiverKt {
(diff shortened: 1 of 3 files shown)


JakeWharton:
This does not leverage any meaningful language feature of Kotlin and sets a precedent for future factory functions that we aren't comfortable with adding. Moreover, creating broadcast receivers isn't something we expect to be done programmatically at a high frequency.

PR #537, opened 2018-05-12
Adding extension for timed logging

RamV13:
The primary benefit of this extension is relieving the burden of calling `reset` and `dumpToLog` before and after each sequence of timed splits (note that `TimingLogger` doesn't provide any other access functions so dumping to the log output is the only use case). Further, this API provides a more declarative approach to mapping split names to the actual work bodies.

**Before**
```kotlin
val timings = TimingLogger(TAG, "methodA")
// ...
timings.reset()
// ... do some work A ...
timings.addSplit("work A")
// ... do some work B ...
timings.addSplit("work B")
// ... do some work C ...
timings.addSplit("work C")
timings.dumpToLog()
```
**After**
```kotlin
val timings = TimingLogger(TAG, "methodA")
// ...
timings.log {
    split("work A") {
        // ... do some work A ...
    }
    split("work B") {
        // ... do some work B ...
    }
    split("work C") {
        // ... do some work C ...
    }
}
```

### DIFF
--- a/api/current.txt
+++ b/api/current.txt
@@ -604,7 +604,8 @@ package androidx.core.util {
 
   public final class TimingLoggerKt {
     ctor public TimingLoggerKt();
-    method public static void log(android.util.TimingLogger, kotlin.Pair<java.lang.String,? extends kotlin.jvm.functions.Function0<kotlin.Unit>>... splits);
+    method public static void log(android.util.TimingLogger, kotlin.jvm.functions.Function1<? super android.util.TimingLogger,kotlin.Unit> work);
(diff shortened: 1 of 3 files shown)


JakeWharton:
I don't think this API design as-is makes much sense. You create a bunch of pair and vararg allocations where it seems like a zero-overhead set of lambdas could be used.

Something like
```kotlin
val logger = TimingLogger(TAG, "methodA")
logger.log {
  split("workA") { .. }
  split("workB") { .. }
}
```

RamV13:
Sorry I completely missed that. Fixed!

romainguy:
Thanks for your PR, I've run the test locally to confirm it passes.

PR #540, opened 2018-05-14
Adding extensions for ImageDecoder

RamV13:
These extensions add `decodeBitmap` and `decodeDrawable` functions to the `ImageDecoder` API for decoding `ImageDecoder.Source` instances. Further, the `decoder` instance is now passed as the `this` to the lambda to enable calling `ImageDecoder` functions more seamlessly.

**Before**
```kotlin
val src = ImageDecoder.createSource(/* ... */)
val b = ImageDecoder.decodeBitmap(src) { decoder, info, source ->
    decoder.setTargetSize(width, height)
    decoder.setTargetColorSpace(ColorSpace.get(ColorSpace.Named.SRGB))
    decoder.setPostProcessor { canvas ->
       canvas.drawTest(/* ... */)
       PixelFormat.UNKNOWN
    }
}
```
**After**
```kotlin
val src = ImageDecoder.createSource(/* ... */)
val b = src.decodeBitmap { info, source ->
    setTargetSize(width, height)
    setTargetColorSpace(ColorSpace.get(ColorSpace.Named.SRGB))
    setPostProcessor { canvas ->
       canvas.drawTest(/* ... */)
       PixelFormat.UNKNOWN
    }
}
```

Code snippet adapted from I/O 2018 (https://youtu.be/eMHsnvhcf78?t=1695)


romainguy:
This is quite different from the conversion functions defined on `Bitmap` and `Color`. This is not a conversion, but an expensive decode. All you are effectively saving is the `ImageDecoder.` prefix in the original call (and admittedly having to use the `decoder` receiver inside the lambda). If the extension was called `decode` instead of `toBitmap` I'd be more inclined to accept it.

RamV13:
That makes sense. What would be a better name then, because `decodeDrawable` and `decodeBitmap` wouldn't clearly indicate that it's decoding **to** that type whereas `decodeToDrawable` and `decodeToBitmap` might be too verbose?

RamV13:
Given the wording in the docs for `ImageDecoder.Source` I think it's reasonable to name the functions `decodeBitmap` and `decodeDrawable`. Updated the PR.

src/androidTest/java/androidx/core/graphics/ImageDecoderTest.kt
```diff
@@ -0,0 +1,66 @@
...
+class ImageDecoderTest {
+
+    @Before fun before() {
+        val bStream = ByteArrayOutputStream().apply {
+            Bitmap
+                .createBitmap(1, 1, Bitmap.Config.ARGB_8888)
+                .compress(Bitmap.CompressFormat.JPEG, 100, this)
+        }
+
+        context.openFileOutput(TEST_FILE.name, Context.MODE_PRIVATE).use {
```
romainguy: Could you just use the ByteBuffer variant of ImageDecoder.createSource instead to avoid relying on the file system during testing?

RamV13: Fixed!

romainguy:
Ran tests locally, everything's green, thanks for the contribution!

PR #543, opened 2018-05-15
Adding extensions for AudioEffect

RamV13:
If you're planning on including extensions for the `android.media` package, I'd like to propose the extensions of `doOnControlStatusChange` and `doOnEnableStatusChange` to `AudioEffect`. The advantage here is leveraging the receiver parameter to avoid explicitly naming the `effect` parameter. Further, parameters can even be omitted altogether by simply utilizing the `it` syntax.

**Before**
```kotlin
equalizer.setControlStatusListener { effect, controlGranted ->
    if (controlGranted) {
        effect.enabled = true
    }
}
```
**After**
```kotlin
equalizer.doOnControlStatusChange { controlGranted ->
    if (controlGranted) {
        enabled = true
    }
}
```

\* test convention for waiting on asynchronous callbacks adapted from AOSP


romainguy:
Thanks, I like it but it looks like a more scalable solution would be to use this: https://kotlinlang.org/docs/reference/compiler-plugins.html#sam-with-receiver-compiler-plugin (we would have to annotated Android APIs but it wouldn't require new implementations/tests/etc. for every SAM we want to improve this way).

RamV13:
Makes sense!

Issue #526, opened 2018-05-02
ComponentName extension function for Context


davethomas11:
What do you think about a simple extension for creating `ComponentName` instances using a `Context`?

```kotlin
inline fun <reified T> Context.componentName() = ComponentName(this, T::class.java)
```

JakeWharton:
Should there be an overload which takes a string as well?

davethomas11:
Sure, I don't see why not. Would there be a case where you have a `Component` you don't have the class reference for? Yes, actually if the `Component` is in another application. Though in that case, you would not be using a `Context` from your application to create it.

There must be a use case why that constructor exists. I'm not sure when `ComponentName(context, string)` is generally used, or if it is merely a convenience method. It does result in one less function call since it does not need to call `getName()` on the class to get the `String` value.

Shrug I'd lean towards not including the string constructor unless I am missing a general use case where this would be useful.

HeyAlex:
Find this really usefull. PR #594

PR #555, opened 2018-05-23
Adding helper function for JobScheduler

### LINKED ISSUE #554: JobScheduler helper function
I'd like to propose a helper function for the JobScheduler API which is fairly cumbersome in Kotlin right now. The signature of the function would be
```kotlin
inline fun scheduleJob(context: Context, componentName: ComponentName, jobId: Int = 0, action: JobInfo.Builder.() -> Unit)
```

**Before**
```kotlin
val builder = JobInfo.Builder(jobId++, serviceComponent)

if (wiFiConnectivityRadioButton.isChecked) {
    builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_UNMETERED)
} else if (anyConnectivityRadioButton.isChecked) {
    builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY)
}

(getSystemService(Context.JOB_SCHEDULER_SERVICE) as JobScheduler).schedule(builder.build())
```
**After**
```kotlin
scheduleJob(jobId++, this, serviceComponent) {
    if (wiFiConnectivityRadioButton.isChecked) {
        setRequiredNetworkType(JobInfo.NETWORK_TYPE_UNMETERED)
    } else if (anyConnectivityRadioButton.isChecked) {
        setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY)
    }
}
```

Please let me know if this is a good idea and I'll get a PR ready!

\* code snippet adapted from Google Samples

RamV13:
Resolves #554


src/androidTest/java/androidx/core/app/job/JobSchedulerTest.kt
```diff
@@ -0,0 +1,66 @@
...
+import kotlin.concurrent.withLock
+
+class JobSchedulerTest {
+    @Test fun testScheduleJob() {
+        scheduleJob(context, ComponentName(context, TestService::class.java), 0) {
+            setOverrideDeadline(1)
+        }
+
+        var looperWaitCount = MAX_LOOPER_WAIT_COUNT
+        lock.withLock {
```
JakeWharton: All of this locking seems like it can be replaced with a simple `CountDownLatch(1)` and await with a reasonable timeout. Right now it seems overly complicated for no good reason.

src/androidTest/java/androidx/core/app/job/JobSchedulerTest.kt
```diff
@@ -0,0 +1,66 @@
...
+import android.content.ComponentName
+import android.support.test.InstrumentationRegistry
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.withLock
+
+class JobSchedulerTest {
+    @Test fun testScheduleJob() {
+        scheduleJob(context, ComponentName(context, TestService::class.java), 0) {
```
JakeWharton: Having to use the context twice here is weird. This function call does not read very well. Perhaps we should address #526 first. But beyond that I'm not sure I'm convinced we need to handle scheduling and job creation in a single function.

src/androidTest/java/androidx/core/app/job/JobSchedulerTest.kt
RamV13: Agreed and updated. I was just following the convention I observed in the Android CTS.

src/androidTest/java/androidx/core/app/job/JobSchedulerTest.kt
RamV13: I've updated the API to be
```kotlin
scheduleJob<TestService>(context, jobId) {
  /* ... */
}
```
which I think reads much nicer. Also, #526 won't change the function call in this case. Do you think this version makes it worth it to handle the scheduling and creation in a single function?

rom4ek:
Shouldn't this be an extension function of `Context` then?

RamV13:
Good point ... fixed

src/androidTest/java/androidx/core/content/ContextTest.kt
```diff
@@ -70,4 +74,29 @@ class ContextTest {
             assertTrue(getInt(R.styleable.SampleAttrs_sample, -1) != -1)
         }
     }
+
+    @Test fun testScheduleJob() {
+        context.scheduleJob<TestService>( 0) {
```
rom4ek: Formatting

src/androidTest/java/androidx/core/app/job/JobSchedulerTest.kt
RamV13: I've updated the API again as per @rom4ek's suggestion and also added an upper bound on the type passed to the `ComponentName`. Please check out the **Benefits** section in the PR description to see why I think this would be worth it. Looking forward to hearing your thoughts!

AllanWang:
Shouldn't this return the int that is returned from `schedule`? And in light of that, I'm not sure if this constitutes a feature addition or not, but you could also use safe casting and return `JobScheduler.RESULT_FAILURE` if the service isn't found, which seems to happen a lot.

RamV13:
Fixed

AllanWang:
You can have the get system service and schedule calls in one line. Also note that I'm just a visitor and not a contributor. Those who manage the project may have a different opinion on the safe casting.

RamV13:
Just keeping it on separate lines for formatting

PR #560, opened 2018-05-25
Added ViewPropertyAnimator extensions

ancirja:
Example: 
 progressView.animate()
.alpha(1)
.doOnEnd { 
    // code
}

### DIFF
--- a/src/androidTest/java/androidx/core/animation/ViewPropertyAnimatorTest.kt
+++ b/src/androidTest/java/androidx/core/animation/ViewPropertyAnimatorTest.kt
@@ -1,0 +1,73 @@
+package androidx.core.animation
+
+import android.support.test.InstrumentationRegistry
+import android.support.test.annotation.UiThreadTest
+import android.support.test.filters.SdkSuppress
+import android.support.test.runner.AndroidJUnit4
+import android.view.View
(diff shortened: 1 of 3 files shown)


src/main/java/androidx/core/animation/ViewPropertyAnimator.kt
```diff
@@ -0,0 +1,94 @@
+package androidx.core.animation
```
romainguy: Missing copyright header

src/androidTest/java/androidx/core/animation/ViewPropertyAnimatorTest.kt
```diff
@@ -0,0 +1,73 @@
+package androidx.core.animation
```
romainguy: Missing copyright header

ancirja: Done. I added the copyright header.

src/main/java/androidx/core/animation/ViewPropertyAnimator.kt
ancirja: Done. I added the copyright header.

PR #567, opened 2018-06-07
Added androidx.core.view.ViewPropertyAnimator

MonteCreasor:
Added androidx.core.view.ViewPropertyAnimator based on androidx.core.animation.Animator. Each extension function returns the ViewPropertyAnimator receiver so that they can be used within the existing anddroid.core.view.ViewPropertyAnimator builder pattern. Also, this extension file does not include doOnPause or doOnResume support.

I guess the thing to consider is that you can't chain multiple 'doOn...' calls since the ViewPropertyAnimator only allows a single listener. Maybe you've already considered that and have decided that since each consecutive 'doOn...' call will overwrite the listener of any previous call, this will lead to accidental complexity. Nonetheless, I personally like the convenience of these extensions.

If you want, I could add an additional comment to each function to warn the caller, for example for the doOnEnd extension:

/**
 * Add an action which will be invoked when the animation has ended.
 * 
 * Note that since ViewPropertyAnimator only supports a single listener, only the last "doOn" 
 * call will be invoked. Use the [setListener] extension when more than one callback is desired.
 *
 * @return the [ViewPropertyAnimator] receiver.
 * @see Animator.end
 */

### DIFF
--- a/src/main/java/androidx/core/view/ViewPropertyAnimator.kt
+++ b/src/main/java/androidx/core/view/ViewPropertyAnimator.kt
@@ -1,0 +1,93 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


MonteCreasor:
I've signed it!

romainguy:
Thank you for the PR, this is unfortunately a duplicate of #560.

PR #570, opened 2018-06-07
Typed getTag view extension

aolivafaura:
Avoid castings when get tagged object on view.
I've run updateApi task but no file was modified, and I don't know how to define it manually. Help is appreciated.


aolivafaura:
Signed CLA :)

Issue #568, opened 2018-06-07
Extension method for checking if the bitmap is empty


kevalpatel2106:
I think it would be great if bitmap has an extension for checking if the bitmap is empty or not based on the byte count.

```
if (bitmap.isEmpty()){
   // Did not read bitmap correctly.
}
```

JakeWharton:
Why based on byte count and not size?

kevalpatel2106:
I didn't use size because `createBitmap` doesn't allow to set the bitmap dimensions to zero. So there is no point of checking the size of the bitmap to see if the bitmap is of 0 * 0 size.

PR #571, opened 2018-06-08
Adding extensions to check if the bitmap is empty or not

kevalpatel2106:
Extension functions to check if the bitmap is empty or not using byte counts. See #568 .

## Usage:

```
if (bitmap.isEmpty()){
   // Did not read bitmap correctly.
}
```

#### or 

```
if (bitmap.isNullOrEmpty()){
   // Bitmap is either null or empty
}
```


romainguy:
Thanks for the PR. This however only indicates whether the bitmap was allocated, not whether the bitmap is empty. Moreover, we don't give a way to allocate 0x0 bitmaps, so I'm not even sure in which situations you could get a Bitmap with an allocation count of 0.

kevalpatel2106:
I had an issue once where the server was sending invalid image bitmap size was zero after bitmap was generated from an input stream. These methods can help to check for the correct bitmap size.

PR #573, opened 2018-06-08
Add empty check functions

pablisco:
Extends `Cursor` to provide a method to 

## Usage

    if(cursor.isEmpty()) {
      // we can show an empty screen
    }

    if(cursor.isNotEmpty()) {
      // We got data
    }

### DIFF
--- a/src/androidTest/java/androidx/core/database/CursorTest.kt
+++ b/src/androidTest/java/androidx/core/database/CursorTest.kt
@@ -18,9 +18,7 @@ package androidx.core.database
 
 import android.database.Cursor
 import android.database.MatrixCursor
-import org.junit.Assert.assertArrayEquals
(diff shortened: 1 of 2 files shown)


JakeWharton:
I don't think we want to use `count` because it requires that the query backing the cursor traverse its entire result set. So if you query for a million items but end up only taking 5 calling `count` will traverse over the entire million to return an accurate number.

pablisco:
Good point, would `isBeforeFirst` be a better call? I do hope not many people have 5 million records on their app database (but wouldn't surprise me either hehe)

pablisco:
I was also contemplating creating a `toSequence(Cursor.() -> T) : Sequence<T>` which would may make it easier to use. But I'm not sure if it's out of the scope of ktx as it's kind of a new feature.

PR #579, opened 2018-06-11
Extension property to get/set bitmap from ImageView

burhanrashid52:
Property extension for fetching bitmap directly from `ImageView` and setting bitmap using `setImageBitmap()` method


burhanrashid52:
I signed it!

PR #582, opened 2018-06-12
P final

JakeWharton:
Closes #581

PR #593, opened 2018-07-08
Add the ability to get the parent Activity of a View

gahfy:
Adding the ability to get the parent Activity of a View.


JakeWharton:
Why do you need this? Accessing the activity from a view is a generally discouraged pattern as it creates a bad coupling that prevents reuse.

gahfy:
@JakeWharton Most of the time, I use it in BindingAdapters in order to put the Observer process out of the Activity to avoid boilerplate code in Activity which is often the same. For example:

```kotlin
/**
 * Sets a mutable visibility to the View. When the value of the LiveData changes, the visibility of
 * the View will change as well.
 * @param view the View to which to set the mutableVisibility
 * @param visibility the visibility LiveData the View should observe
 */
@BindingAdapter("mutableVisibility")
fun setMutableVisibility(view: View, visibility: MutableLiveData<Int>?) {
    val parentActivity: AppCompatActivity? = view.getParentActivity()
    if(parentActivity != null && visibility != null) {
        visibility.observe(parentActivity, Observer { value -> view.visibility = value?: View.VISIBLE})
    }
}
```

Then all you have to do is to set `app:mutableVisibility` as attribute in XML, and you don't have to care about Observers in your Kotlin `Activity`.

gahfy:
Rebased on #582 as it has been merged to let CI pass

gahfy:
Does someone understand the error of instrumented tests and how to solve it?

